PR #64, opened 2017-04-24
creation of model training recipes (exa, zero shot and one-model multilingual)

DevSinghSachan:
* added the recipe directory in which model config files can example scripts can be kept.
* added an example of how to read files from TED talks corpus and further tokenize it using installed google's sentencepiece code.
* Added example of zero shot learning for pt-br and es using en as indirect language as in Google Multilingual paper.


neubig:
I'm going to close this for now. Feel free to reopen if we get a new recipe that matches the new config files, etc.

PR #75, opened 2017-05-25
add optional arguments to pass to dynet

john-hewitt:
A couple of common dynet flags  (`--dynet-mem` and `--dynet-gpu-ids`) can be necessary for running xnmt.
As of now, both flags cause argparse to throw an error in both xnmt_train and xnmt_decode. `--dynet-gpu-ids` also causes argparse to crash in xnmt_run_experiment.

This commit adds the flags as optional arguments so they can be passed transparently to dynet.


neubig:
Thanks!

PR #79, opened 2017-05-29
Fix serialization

msperber:


msperber:
Some changes that were necessary to get model serialization / deserialization working for me (debug config extended to test model loading).
It would be nicer to not have to create all the little classes in encoder.py, but I would need a better understanding of how dynet RNN serialization works to do that. I think it's better not to return a Builder class as an encoder, because we might add general-purpose features to encoders (like set_train(bool) which would be an important feature).

neubig:
Thanks, and sorry about breaking things! I was just looking at this and re-discovered how much we need unit tests. I'll try to create some today.

Regarding not creating all the little classes in `encoder.py`, I'll take a stab at that after merging this commit.

PR #82, opened 2017-05-29
Train test behavior

msperber:
Suggestion for how to separate train vs test behavior, which should be useful for dropout, batch normalization etc.

Included sample dropout implementation for bilstm encoder.


neubig:
Thanks! I really like how you did this, and just have a few comments:
* The naming `TrainTestBehavior` threw me off a little bit. Is `Behavior` the canonical way to call something like this in Python? Java/C# have the word `Interface` which would have been easier for me to understand.
* We don't need getters in Python, you can just set `train_test_components`, and we can iterate over them.

In fact, now that I think about it, we might not need an Behavior/Interface at all. What if we just have a function in Translator (which will always be the top of our tree of modules) iterate over train_test_components (maybe recursively if some of the values in its train_test_components have their own train_test_components)? This might simplify things.

msperber:
Thanks for the comments! Yeah, "behavior" is a little strange / non-standard. I don't think there's any Python conventions other than being consistent. I renamed it to interface. And making the translator responsible for calling all sub components recursively sounds good!

We probably shouldn't get rid of the interface though, because some classes (probably the leaves in this hierarchy) actually have to implement how to switch between training / testing mode. And I thought since we already have an interface, it would make things easier to understand if we are explicit about how to specify sub_components. I think it's fairly common to provide a getter method in such cases. What do you think?

neubig:
OK looks good, but this caused a merge conflict, sorry. Go ahead and merge once you can fix it!

PR #83, opened 2017-06-01
Encoder spec via dictionaries

msperber:
This is to allow parametrizing encoders arbitrarily from the config files.
We can certainly do more refactoring, but it's ready to use so I thought I'd throw it out now.

One question is whether and how to document encoder parameters. There's 3 parameters that every encoder has: type, input_dim, output_dim. The encoder-specific parameters are kind of formalized in BuilderEncoder.use_params(), we could think about developing this more fully and even creating documentation automatically. I need to take care of other things now, but I'll be making heavy use of this and probably be pushing fixes/updates.

### DIFF
--- a/examples/debug.yaml
+++ b/examples/debug.yaml
@@ -21,7 +21,8 @@ defaults:
 
 debug-config-1layer:
   train:
-    encoder_layers: 1
+    encoder:
+      layers: 2
 
 debug-config-2layers:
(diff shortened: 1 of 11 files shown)


neubig:
Thanks, this is great! I think it's fine to merge for now, and we can make any small adjustments later.

PR #84, opened 2017-06-02
setting decoder RNN input dim to size of trg embeddings, not of encoder output

msperber:


neubig:
Thanks, this looks fine to merge once conflicts are resolved.

PR #89, opened 2017-06-08
Updated experiment_config_files.rst

roeeaharoni:
Updated experiment_config_files.rst to match current implementation.


neubig:
Thanks!

PR #92, opened 2017-06-14
decoupled reporting of train / dev loss

msperber:
behavior is now that train PPLs are reported every 1k sents, and only dev reporting (incl. model saving and LR decay) is configurable via "eval_every"


neubig:
Thanks!

PR #95, opened 2017-06-16
pyramidal lstm: select skip or concat mode

msperber:


neubig:
LGTM, but now there are merge errors. Go ahead and merge once they're fixed.

PR #99, opened 2017-06-23
Random search

msperber:
Feature that performs random search over model / search parameters, by simply specifying the choices in the config file. Search can be sequential or in parallel, depending on how experiment runner is started.

### DIFF
--- a/examples/random-search.yaml
+++ b/examples/random-search.yaml
@@ -1,28 +1,34 @@
 # Demonstrates random search over model parameters
 defaults:
   experiment:
-    model_file: ../examples/output/<EXP>.mod
-    hyp_file: ../examples/output/<EXP>.hyp
-    out_file: ../examples/output/<EXP>.out
-    err_file: ../examples/output/<EXP>.err
+    model_file: examples/output/<EXP>.mod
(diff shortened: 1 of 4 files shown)


neubig:
Nice idea. One simple question though, what if we want the encoder and decoder to have the same number of layers, but random? Is there any way to accommodate this?

msperber:
That should be possible by doing random search over default_layer_dim (maybe manually fixing the other related parameters if the search should be only over encoder / decoder dim).

Otherwise, one could also try named random parameters and then somehow refer to that (it's possible that YAML supports something like that), but I'm not sure if that would make things too complicated?

neubig:
Another option would be to make it possible for parameters to be references to other parameters (e.g. decoder_dim = encoder_dim), but this also seems a little bit complicated. One option is to just leave things as-is for now, and then fix this when it becomes a problem. If you agree, go ahead and merge :)

msperber:
It was actually relatively easy, so now you can do something like:
    encoder:
      hidden_dim: !RandomParam
        values: [32,64]
    attention_context_dim: !RefParam
      ref: encoder.hidden_dim

neubig:
Great, go ahead and merge this.

PR #100, opened 2017-06-23
Partial length norm

msperber:
Some variants of length normalization may be applied already during the beam search rather than after the search, which may produce better results.


neubig:
Thanks!

PR #102, opened 2017-06-26
refactored encoder module

msperber:
Code is a bit cleaner now. A last remaining refactoring step might be to collapse related encoders that have the same parameters into one encoder class.

### DIFF
--- a/examples/debug.yaml
+++ b/examples/debug.yaml
@@ -16,7 +16,7 @@ defaults:
     default_layer_dim: 64
     dropout: 0.5
     encoder:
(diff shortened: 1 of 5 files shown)


examples/debug.yaml
```diff
@@ -16,6 +16,7 @@ defaults:
     default_layer_dim: 64
     dropout: 0.5
     encoder:
+      type: PyramidalLSTM
```
neubig: It's a bit strange to have the default be PyramidalLSTM here.

msperber:
True, I fixed it!

PR #103, opened 2017-06-26
Consistently using/enforcing unicode; fix BPE postprocessing

msperber:


neubig:
This is great! I'm actually thinking of unifying all pre/post processing as part of the Normalizer code at some point to enforce consistency across all uses of pre/post processing. For the time being I think it's fine to merge this though.

PR #104, opened 2017-06-26
Add preprocessing capability

neubig:
This commit makes it possible to perform some rudimentary preprocessing directly within xnmt. There are still several missing components, which will be added in following commits:

1) Tokenization/detokenization
2) Separate specification of what preprocessing should be done before evaluation

### DIFF
--- a/examples/preproc.yaml
+++ b/examples/preproc.yaml
@@ -19,36 +19,34 @@ defaults:
       - examples/data/test.ja
       - examples/data/test.en
       out_files:
-      - examples/out/train.norm.ja
-      - examples/out/train.norm.en
-      - examples/out/dev.norm.ja
-      - examples/out/dev.norm.en
-      - examples/out/test.norm.ja
-      - examples/out/test.norm.en
+      - examples/output/train.norm.ja
+      - examples/output/train.norm.en
+      - examples/output/dev.norm.ja
+      - examples/output/dev.norm.en
+      - examples/output/test.norm.ja
+      - examples/output/test.norm.en
       specs:
       - filenum: all
         spec:
         - type: lower
     - type: filter
       in_files:
-      - examples/out/train.norm.ja
-      - examples/out/train.norm.en
+      - examples/output/train.norm.ja
+      - examples/output/train.norm.en
       out_files:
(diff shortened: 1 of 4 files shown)


msperber:
Looks basically good to me! My main comment is already a TODO in your code, which would be handling of non-text inputs. That would mostly concern the filtering, not the normalization and vocab operations. For the filtering, how about passing a SentenceFilterer to the training step, and applying it while reading in the files in [here](https://github.com/neulab/xnmt/blob/master/xnmt/xnmt_train.py#L174)? That might be a scalable way of supporting filtering for arbitrary input readers.

Other minor comments:
- should things like splitting into characters or BPE units be a TODO to be integrated into the preprocessing?
- for the [suboptions](https://github.com/neulab/xnmt/blob/add-preproc/xnmt/xnmt_preproc.py#L12), I implemented the "." notation for use with the encoders, see [here](https://github.com/neulab/xnmt/blob/add-preproc/xnmt/xnmt_train.py#L55). I'm not sure if it's usable without change since you have a list in between, but it might be straightforward to use. It's basically what [collapse_sub_dict](https://github.com/neulab/xnmt/blob/add-preproc/xnmt/xnmt_train.py#L55) and [update_with_subdicts](https://github.com/neulab/xnmt/blob/add-preproc/xnmt/options.py#L94) are used for.

msperber:
By the way, part of the reason why I mentioned doing the filtering while loading the data is that for non-text data, reading / writing data can involve a significant amount of time and disc space, and we could avoid one reading+writing pass in this way.

neubig:
For 1, sure this sounds like a good idea. I think it should be an option in the preprocessor, but we could also make it an option in the trainer as well. I also expect to add normalization as an option to the decoder as well, for the case where we might want to perform end-to-end translation from raw text. If you want to add this now I think that'd be great, otherwise was can save it for a future commit.

For 2, yeah, that's part of the plan. I made this explicit in #107 

For 3, I agree that the implementation now is probably not the best way, but wasn't able to figure out how to do so. If you could modify this to be more canonical I'd be happy to go with that.

msperber:
Okay, sounds good! I guess we can merge now and add these features later.

PR #118, opened 2017-07-05
Optimization / learning schedule

msperber:
- optionally restart Adam when applying LR decay
- revert to best checkpoint when applying LR decay
- specify num attempts before LR decay
- determine early stopping / LR decay based on BLEU and other criteria
- eval test data only at end of experiment

### DIFF
--- a/examples/standard.yaml
+++ b/examples/standard.yaml
@@ -36,7 +36,6 @@ defaults:
         layers: 1
         mlp_hidden_dim: 512
   decode:
-    post_process: join-char
     src_file: examples/data/test.ja
   evaluate:
     ref_file: examples/data/test.en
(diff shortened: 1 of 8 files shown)


neubig:
Awesome, thanks! It would also be nice if in the future we could have a "patience" where it waits for N dev evaluations to produce sub optimal results, then reloads the model that got the best score on the dev set and restarts training with a lower rate. This is what we used in our model on stronger baselines for neural MT

msperber:
Yeah, I intended to do that with the attempts_before_lr_decay option. Is that what you were thinking of? I can rename it to patience, that sounds better..

neubig:
Ahh, perfect. I just missed that option. Thanks!

PR #122, opened 2017-07-07
Sample from data @ each epoch

msperber:
Useful to train on large (especially non-text) data without having to load everything into memory, as well as to bring more randomness into the training.

### DIFF
--- a/xnmt/input.py
+++ b/xnmt/input.py
@@ -200,17 +200,21 @@ class BilingualCorpusParser(CorpusParser, Serializable):
     self.max_num_train_sents = max_num_train_sents
     self.max_num_dev_sents = max_num_dev_sents
     self.sample_train_sents = sample_train_sents
+    self.train_src_len, self.train_trg_len = None, None
+    self.dev_src_len, self.dev_trg_len = None, None
     if max_num_train_sents is not None and sample_train_sents is not None: raise RuntimeError("max_num_train_sents and sample_train_sents are mutually exclusive!")
 
   def read_training_corpus(self, training_corpus):
     training_corpus.train_src_data = []
     training_corpus.train_trg_data = []
-    src_len = self.src_reader.count_sents(training_corpus.train_src)
-    trg_len = self.trg_reader.count_sents(training_corpus.train_trg)
-    if src_len != trg_len: raise RuntimeError("training src sentences don't match target sentences: %s != %s!" % (src_len, trg_len))
+    if self.train_src_len is None:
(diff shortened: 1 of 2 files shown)


xnmt/input.py
```diff
@@ -114,16 +141,22 @@ class ContVecReader(InputReader, Serializable):
   def __init__(self):
     pass
 
-  def read_sents(self, filename, max_num=None):
+  def read_sents(self, filename, filter_ids=None):
     npzFile = np.load(filename)
```
neubig: Does this end up loading the whole corpus into memory? If so, maybe we need to add `mmap_mode="r"`?

xnmt/input.py
```diff
@@ -159,19 +183,41 @@ class BilingualCorpusParser(CorpusParser, Serializable):
...
     self.max_num_dev_sents = max_num_dev_sents
+    self.sample_train_sents = sample_train_sents
+    if max_num_train_sents is not None and sample_train_sents is not None: raise RuntimeError("max_num_train_sents and sample_train_sents are mutually exclusive!")
 
   def read_training_corpus(self, training_corpus):
     training_corpus.train_src_data = []
     training_corpus.train_trg_data = []
-    for src_sent, trg_sent in six.moves.zip(self.src_reader.read_sents(training_corpus.train_src, self.max_num_train_sents),
-                                            self.trg_reader.read_sents(training_corpus.train_trg, self.max_num_train_sents)):
+    src_len = self.src_reader.count_sents(training_corpus.train_src)
```
neubig: This will result in always iterating over the corpus twice, right? I think you should only do this when necessary (e.g. when using `sample_train_sents`). It would be even better if we only had to do it once at the beginning of training (e.g. cache it in something like `training_corpus.train_src_len` and re-use the cached value).

xnmt/input.py
msperber: Yes, thanks for catching this!

xnmt/input.py
msperber: Yes, it goes over it twice, although count_sents should be faster than read_sents. I added the caching for better speed. The reason I computed src_len / trg_len even without sampling is that zip doesn't fail when the iterators have different lengths, and I thought it would be good to have that as a bug check. There might also be a better solution for this..

neubig:
Re: checking the length, one option would be to use `itertools.zip_longest()`, then check for `None` every time. I'm not sure which is better though, so I'll leave the decision up to you. If you think it's fine as-is, go ahead and commit!

PR #126, opened 2017-07-10
lazy expression sequence to avoid unnecessary pick() operations

msperber:
Currently, when using contvec inputs and RNN encoders, the contvec would produce a tensor expression sequence, and the RNN would use pick() for each input. LazyNumpyExpressionSequence helps avoid those unnecessary operations.


neubig:
Good idea, thanks!

PR #129, opened 2017-07-11
fixed python2 when writing non-unicode logs

msperber:
(got broken by the recent python3 fix)


neubig:
@philip30 Woops... We really need to get continuous integration working so it's easier to test these things. Could you fix this bug?

philip30:
I think using six.u is the solution. It will return str in python3 and unicode in python2.

PR #130, opened 2017-07-11
Fixes some python2-3 compatibilities

philip30:
Haven't checked in python2, but I am pretty sure it works.

Btw, we actually need to always use six.moves.map that returns generator as list(map()) expression is 2 times N operations in python2.


neubig:
Thanks!

msperber:
Thanks, and sorry for breaking things again!

PR #131, opened 2017-07-12
Simplified batcher and fixed bug

neubig:
This commit simplified batcher. I also noticed that there was probably a bug where a minority of sentences (particularly when using very small data) were becoming unaligned, and this behavior was inconsistent. I'm not sure why this was happening in the previous code, but I think this commit should fix it.

### DIFF
--- a/examples/debug.yaml
+++ b/examples/debug.yaml
@@ -10,7 +10,7 @@ defaults:
   train:
     default_layer_dim: 64
     dropout: 0.5
-    dev_metrics: bleu    
+    dev_metrics: bleu
     training_corpus: !BilingualTrainingCorpus
       train_src: examples/data/head.ja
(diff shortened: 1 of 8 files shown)


msperber:
It's looking good to me basically! I do have 2 questions:
- It looks like the word batcher has been dropped, is there a particular reason for that? I found it quite useful when working close to the memory limit, so I'm wondering if it would make sense to keep it.
- There is no shuffling of batches as far as I can see. Shuffling the batches probably should be done by the training loop and not the batcher anyways, maybe now is a good chance to add that? Previously, I think the batches were only shuffled once before the first epoch when the batches created which probably wasn't ideal.

neubig:
OK, both of these should be fixed now.

PR #134, opened 2017-07-13
Removed extra whitespaces from xnmt

philip30:
As you can see, I am removing extra whitespaces from the XNMT source code. I think github can detect it. I also added a bash script to remove the whitespace from the code under the script/code/clean.sh for easy future use.

My editor can detect the extra spaces, so it is a bit bothering to see red everywhere. I think it is a good practice also.


neubig:
Thanks!

PR #135, opened 2017-07-13
Cont Vector Reader should be able to read vector shape that is not assumed by the reader.

philip30:
Currently, I am adding the ability to transpose in the reader. Apparently, the dataset from the speech has the format that is transposed from what the reader assumed.


neubig:
Good idea, thanks!

PR #137, opened 2017-07-17
Added ability to create reports for translations

neubig:
This commit makes it possible to create HTML reports for each translated sentence, including attention visualization by @shrutijpalaskar . This is not written in a super-extensible way at the moment, but I plan on stress-testing the extensibility by trying to use this for speech-to-text translation, or visualizing layers of the network, etc. This will be added in a future pull request.

### DIFF
--- a/examples/debug-report.yaml
+++ b/examples/debug-report.yaml
@@ -1,0 +1,47 @@
+# Small config to help refactoring to make the model completely YAML based
+defaults:
+  experiment:
+    model_file: examples/output/<EXP>.mod
+    hyp_file: examples/output/<EXP>.hyp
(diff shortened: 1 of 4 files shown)


neubig:
I think that'd probably be one of the next things on the plate! @shrutijpalaskar , if you think it wouldn't be too hard to create a thing to create a similar heatmap for numpy arrays where the rows represent speech features and columns represent time, that would be another excellent addition. (An example of some data that could be used to do this is the `.npz` files here: https://github.com/neulab/xnmt/tree/master/examples/data)

shrutijpalaskar:
Hi all, I'm looking into getting a attention visualization for speech-to-text next. Shall have something soon!

PR #138, opened 2017-07-17
deserialization fix for python 2.7

msperber:
six.u() only works for string/unicode inputs


neubig:
Hmm, I actually thought I had fixed this already... But thanks!

PR #139, opened 2017-07-18
Add tokenization: sentencepiece and arbitrary external wrapper

john-hewitt:
### Implemented in this PR:
 - Ability to run external tokenizers on entire files at once, to reduce external executable overhead, or on a per-sentence level, for flexibility.
 - Tokenization pipeline that never writes any file to disk until the very last tokenizer to reduce I/O.
 - `ExternalTokenizer` class for specifying arbitrary tokenizer executable and passing options to it, along with in_files and out_files.
 -  `SentencepieceTokenizer` class for useful syntactic sugar in specifying sentencepiece as an external tokenizer.
- `xnmt_preproc.py` tokenization step.
- `preprocessing.rst` documentation of current tokenization capabilities.


msperber:
It's looking good from my side! One comment would be that we're also trying to support computing eval metrics during and after training, which would require some of the tokenization to be undone. For example, we would probably undo BPE but not Moses-tokenization. Currently we have the postprocess option that can merge characters or BPE, but it's probably not flexible enough. One way might be to include an undo_tokenization() method for each tokenizer, and then find some place where we can specify which steps should be undone for scoring?

neubig:
Thanks, this looks good in general as well! A couple comments:

* Since I wrote the preprocessing functionality, Matthias has done some restructuring of the YAML file to make it more elegant, specifying a class name such as `!LSTMEncoder` directly in the YAML file, then having the constructor of that class be called automatically. Specifically, you can take a look at how the encoders are implemented. I'd suggest going this route when creating your tokenizers (I'll also try to do this for the other pre-processing steps as well).
* Any external scripts that you copy into xnmt should be put in a separate directory (maybe "third_party"). This is especially true for things from Moses, which is a different license than xnmt.
* BPE and sentencepiece are trainable, so you'll need to make sure that they can be trained, but I imagine that you're thinking of this already.

john-hewitt:
@msperber  I'm not entirely sure what that will look like yet, with the variation in tokenizers that we might see. We could include, in each tokenizer constructor, a flag  `should_undo`, or like, that would decide whether such an `undo_tokenization()` would be a no-op, or actually undo the tokenization.

@neubig Unfortunately, pyyaml.org was down for a bit this weekend, which was frustrating. But the latest commits use the new yaml syntax.

I agree on the separation of out-of-license material; I'd go as far as to include something like a `download_third_party.sh` file that prints some warning message about licensing before it downloads the moses tokenizer or like.

Regarding training -- I've decided to include a list of paths to train files in the `BPE` constructor. It duplicates paths with `in_files`, but is the cleanest option I considered.

The commits so far still don't have real functionality.

john-hewitt:
Sentencepiece has been integrated nicely (as long as you can point to your installation.)
Any old external tokenizer will also work.

TODO is a native BPE implementation as well as the ability to let Sentencepiece refer to files tokenized by earlier steps in the process. Right now it can only reference original, untokenized files. This latter item I plan to fix soon, but this works as is for individual tokenization schemes, so it might be worth separate commits.

doc/source/preprocessing.rst
```diff
@@ -0,0 +1,58 @@
...
+                          to the experiment config.
+                          See the Sentencepiece section for more specific information on this tokenizer.
+
+2. External Tokenizers:   Any external tokenizer can be used as long as it tokenizes stdin and outputs
+                          to stdout. A single Yaml dictionary labelled ``tokenizer_args``
+                          is used to pass all (and any) options to the external tokenizer.
+                          The option ``detokenizer_path``, and its option dictionary, ``detokenizer_args``,
+                          can optionally be used to specify a detokenizer.
+
+3. Byte-Pair Encoding:    A compression-inspired unsupervised sub-word unit encoding
```
neubig: Comment this out until it's finished?

examples/preproc.yaml
```diff
@@ -19,6 +19,33 @@ defaults:
...
+      - examples/data/train.tok.en
+      - examples/data/dev.tok.ja
+      - examples/data/dev.tok.en
+      - examples/data/test.tok.ja
+      - examples/data/test.tok.en
+      specs:
+      - filenum: all
+        tokenizers:
+        - !ExternalTokenizer
+          path: /nlp/users/johnhew/nmt/xnmt/script/code/toy_word_tokenize.py
```
neubig: Change to relative path, or if it should be absolute, make it generic.

neubig:
Thanks, this is great. I think it's ok to go ahead and integrate this now, then add the additional functionality later.

examples/preproc.yaml
john-hewitt: Yep, oversight on my part.

doc/source/preprocessing.rst
john-hewitt: Makes sense; done.

neubig:
Thanks!

PR #140, opened 2017-07-18
fixed learning rate, moved to new-style classes

msperber:
I hope the handling of the learning rate is correct according to the new DyNet conventions?

### DIFF
--- a/xnmt/xnmt_train.py
+++ b/xnmt/xnmt_train.py
@@ -56,7 +56,7 @@ options = [
   Option("model", dict, default_value={}),
 ]
 
-class XnmtTrainer:
+class XnmtTrainer(object):
   def __init__(self, args, output=None):
     dy.renew_cg()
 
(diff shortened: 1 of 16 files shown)


xnmt/xnmt_train.py
```diff
@@ -112,11 +112,12 @@ def pack_batches(self):
 
   def dynet_trainer_for_args(self, args):
     if args.trainer.lower() == "sgd":
-      trainer = dy.SimpleSGDTrainer(model_globals.dynet_param_collection.param_col, learning_rate = args.learning_rate)
+      trainer = dy.SimpleSGDTrainer(model_globals.dynet_param_collection.param_col)
```
neubig: It's still fine to pass the learning rate through the constructor. Was this not working?

msperber: Hm, I got an error because learning_rate was not a known parameter. I did use the recent DyNet, but maybe there was still a versioning problem. In either case, if we're at it I guess we could also change this to support any trainer that's available in DyNet by using getattr()..

PR #142, opened 2017-07-19
Made packer re-pack batches at every epoch when shuffling

neubig:
When choice of mini-batches plays a particularly large role in the stability of the training, it might be good to ensure different mini-batches on each epoch. This commit makes this possible by ensuring that the batches are re-created every epoch when using a batcher that includes randomness.


neubig:
Yeah, I think that'd be a good idea.

PR #144, opened 2017-07-19
Speech retrieval to merge

neubig:
We've bene working a bit on a separate "speech_retrieval" branch, but I think things have stabilized a bit and we can merge into the main branch.


neubig:
@lwang114 @DannyMerkx @philip30 @mingxingdu I did some refactoring and am going to merge this into the master branch shortly. After this, I think we can merge this into the speech retrieval branch (I'll do this), and continue development. Big changes are:

* the Harwath encoders are moved to the specialized_encoders file
* the continuous vector input no longer transposes by default

These might break experiments.

PR #145, opened 2017-07-19
Make it possible to measure recall on only a subsample of the database IDs

neubig:
In retrieval tasks, it's common to only measure retrieval accuracy over the elements in the test database. This commit makes it possible to measure accuracy in this way.


philip30:
LGTM!

PR #146, opened 2017-07-20
make argument explicitly optional because default is None

john-hewitt:
Quick matter I picked up while trying to grok the option system:

The example configs are broken right now by missing `candidate_id_file`, an option that seems to be intended to be optional, but is required. 

Specifically, due to [the option parsing at this line](https://github.com/neulab/xnmt/blob/master/xnmt/options.py#L27), an option is required if the `required` flag is set to `True` or if the `required` flag is set to `None`, and the default value is `None`. 

I did not want to accidentally break functionality by changing the default value of `candidate_id_file`, so I made it optional.


neubig:
Thanks! I just missed the optional flag.

PR #147, opened 2017-07-26
Refactor XNMT

philip30:
There are several major refactoring I made and I propose and a new design for XNMT.
- Introduced HierarchicalModel. This class is a representation of a hierarchical data structure object. We add the direct descendant by calling register_hier_child() method in the constructor.
- Added decorators 'recursive': We add this decorator to the method of the subclass of HierarchicalModel. Then, if the method is called in the parent, the direct descendant will also invoke the method recursively. Because of the recursion, this is a pre-order traversal of method callings. For example, we can use this on set_train, because there are several descendants who need to implement the logic of this method. Consequently, the TrainTestInterface is not needed because it is redundant (the recursive call there is already suported by this decorator).
- Removed static methods from Batcher. I don't think it is good to call a static method every where. So I moved it into a module level. It is still the same, but not bounded by the class.
- Introduced Generator class who is the superclass of Translator and Retriever.
- Introduced lxml as html builder (need to pip install it!). Together with the 'recursive', we can smartly modify the inner element of html tree when the children inherit the HTMLReportable class.
- Removed most of the report class and its implementation and moved the logic to the class who is responsible for generating the html document.
- I removed some of the * imports. I think it is not a good practice as it makes difficult to trace the source code of the class (and can cause more severe problems).


xnmt/retriever.py
```diff
@@ -30,11 +33,11 @@ def __init__(self, reader, database_file, dev_id_file=None, test_id_file=None):
     self.test_id_file = test_id_file
 
   def __getitem__(self, indices):
-    return Batcher.mark_as_batch(Batcher.pad([self.data[index] for index in indices]))
+    return batcher.mark_as_batch(batcher.pad([self.data[index] for index in indices]))
 
 ##### The actual retriever class
 
-class Retriever(TrainTestInterface):
+class Retriever(GeneratorModel, HierarchicalModel):
```
msperber: Why do we need to subclass both, although GeneratorModel is already a subclass of HierarchicalModel?

xnmt/encoder.py
```diff
@@ -95,8 +113,8 @@ def transduce(self, sent):
       sent = module.transduce(sent)
     return sent
 
-  def get_train_test_components(self):
-    return self.modules
-
-
+  @decorators.recursive
```
msperber: Couldn't we handle this also with recursion instead of explicit code?

xnmt/model.py
```diff
@@ -0,0 +1,34 @@
...
+      generation_output = self.post_processor.process_outputs(generation_output)[0]
+    return generation_output
+
+  def generate(self, *args, **kwargs):
+    raise NotImplementedError()
+
+  def initialize(self, system_args):
+    pass
+
+  def set_vocabs(self, src_vocab, trg_vocab):
```
msperber: This seems a bit inflexible, for example when encoding speech we have no src_vocab.

philip30: It is true. But it is actually hard if the translator does not have an access to vocabulary and wants to generate and HTML based report inside. How about if we generalize it into add_resource?

xnmt/encoder.py
philip30: There will be many tedious recursion code because, for example the HTMLReportable need to recurse through all the members that implements the method html_report. By marking the code with this decorator, it does that automatically.

xnmt/retriever.py
philip30: I will remove it. Thank you for catching!

xnmt/model.py
philip30: I remove the vocabs and let them be the arguments of generate_html_report instead

philip30: easier version: I just move the method to the Translator class

xnmt/decorators.py
```diff
@@ -0,0 +1,26 @@
+import model
+
+def recursive(f):
```
philip30: Done!

philip30:
@msperber, @neubig because the comments are resolved and I want to move forward with other things, I am merging this to master. Any following bad side effects will be fixed by hotfix push.

PR #149, opened 2017-07-31
add linear_encoder

lwang114:
Create a more general encoder class called linear_encoder to replace the too specific HarwathImageEncoder, which is essentially a linear projection.

### DIFF
--- a/xnmt/encoder.py
+++ b/xnmt/encoder.py
@@ -1,4 +1,5 @@
 import dynet as dy
+import expression_sequence
 import model
 import embedder
 import serializer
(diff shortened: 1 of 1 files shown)


xnmt/encoder.py
```diff
@@ -119,3 +119,45 @@ def set_train(self, val):
...
+    src = src.as_tensor()
+
+    src_height = src.dim()[0][0]
+    src_width = 1
+    batch_size = src.dim()[1]
+
+    W = dy.parameter(self.pW)
+    b = dy.parameter(self.pb)
+
+    src = dy.reshape(src, (src_height, src_width), batch_size=batch_size) # ((276, 80, 3), 1)
```
neubig: Is this reshape necessary?

xnmt/encoder.py
```diff
@@ -119,3 +119,45 @@ def set_train(self, val):
...
+    # convolution and pooling layers
+    l1 = (W*src)+b
```
neubig: Changing to `dy.affine_transform(b, W, src)` will save memory.

Also, perhaps we can add an option to have some sort of non-linearity (`tanh`, `relu`, `sigmoid`)?

PR #151, opened 2017-08-01
add with_bias option and error message

lwang114:
Raise an error message if the nonlinearity is not known to the encoder. Specify with_bias to indicate whether the layer uses bias or not

### DIFF
--- a/xnmt/encoder.py
+++ b/xnmt/encoder.py
@@ -128,7 +128,7 @@ class FullyConnectedEncoder(Encoder, Serializable):
     Then, we add a configurable number of bidirectional RNN layers on top.
     """
 
-  def __init__(self, in_height, out_height, nonlinearity='linear'):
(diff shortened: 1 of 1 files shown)


xnmt/encoder.py
```diff
@@ -159,6 +159,8 @@ def transduce(self, src):
     # convolution and pooling layers
     #l1 = (W*src)+b
     l1 = dy.affine_transform([b, W, src])
```
neubig: This should be inside `else` following `if` in the below line, right?

xnmt/encoder.py
```diff
@@ -159,6 +159,8 @@ def transduce(self, src):
...
+    if self.with_bias is False:
+      l1 = W * src
     output = l1
     if self.nonlinearity is 'linear':
```
neubig: Instead of:

    if x:
    else:
      if y:
      else:

you should use

    if x:
    elif y:
    else:

PR #152, opened 2017-08-01
Segmenting encoder

philip30:
Before it diverges too far, I want to merge this branch into master.
Basically several new features:

1. Added LossBuilder where we can compose losses from several methods
2. Added the calc_reinforce_loss to retriever and translator
3. Added Linear class as a wrapper for affine_transform.
4. Added SegmentingEncoder (+test_segmenting.yaml)


examples/retrieval.yaml
```diff
@@ -14,7 +14,7 @@ defaults:
     dropout: 0.0
     dev_metrics: recall|nbest=1,recall|nbest=5
     training_corpus: !BilingualTrainingCorpus
-      train_src: examples/data/train.ja
+      train_src: examples/data/train-char.ja
```
neubig: Revert this.

examples/retrieval.yaml
```diff
@@ -45,4 +45,4 @@ defaults:
 
 standard-dropout0.1:
   train:
```
neubig: You can just remove the whole `train:` block and rename `standard-dropout0.1` to `retrieval`.

xnmt/embedder.py
```diff
@@ -7,6 +7,14 @@
 from serializer import Serializable
 from expression_sequence import ExpressionSequence, LazyNumpyExpressionSequence
 
+class Embedding(object):
```
neubig: Why is this necessary?

xnmt/encoder.py
```diff
@@ -20,19 +23,37 @@ class Encoder(HierarchicalModel):
...
+      raise NotImplementedError("Unimplemented transduce logic for class:",
+                                self.builder.__class__.__name__)
+
+    return ExpressionSequence(expr_list=out)
+
+class IdentityEncoder(Encoder, Serializable):
+  yaml_tag = u'!IdentityEncoder'
+
+  def transduce(self, embed_sent):
+    return ExpressionSequence(expr_list = sent.get())
```
neubig: This code should just return `embed_sent`, right? Encoders assume they receive an expression sequence and return an expression sequence.

xnmt/encoder.py
```diff
@@ -146,8 +209,8 @@ def __init__(self, in_height, out_height, nonlinearity='linear'):
     self.pW = model.add_parameters(dim = (self.out_height, self.in_height), init=normalInit)
     self.pb = model.add_parameters(dim = self.out_height)
 
-  def transduce(self, src):
-    src = src.as_tensor()
+  def transduce(self, embed_sent):
+    src = embed_sent.get().as_tensor()
```
neubig: This doesn't match the current API. Why the `get()` function?

xnmt/input.py
```diff
@@ -173,13 +181,54 @@ def count_sents(self, filename):
     npzFile.close()
     return l
 
+class ContVecNPYReader(InputReader, Serializable):
```
neubig: Remove this.

xnmt/input.py
```diff
@@ -173,13 +181,54 @@ def count_sents(self, filename):
...
+    return l
+
 class IDReader(BaseTextReader, Serializable):
   """
   Handles the case where we need to read in a single ID (like retrieval problems)
   """
   yaml_tag = u"!IDReader"
 
   def read_sents(self, filename, filter_ids=None):
+    #print(filename)
```
neubig: And this.

xnmt/translator.py
```diff
@@ -37,17 +37,11 @@ def calc_loss(self, src, trg):
     '''
     raise NotImplementedError('calc_loss must be implemented for Translator subclasses')
 
-  def translate(self, src):
-    '''Translate a particular sentence.
-
-    :param src: The source, a sentence or a batch of sentences.
-    :returns: A translated expression.
+  def calc_reinforce_loss(self, reward):
```
neubig: Can we come up with a more general name? We might not necessarily be using reinforce, but still want to use this capability.

philip30:
Thanks for the review @neubig! 

Yes, I agree that it will break the API, thanks for pointing that out.
Actually the reason I did that is because the current embedder will strip out some useful information / annotation that are produced by the reader.

To sum this up.
Embedder: list(SimpleSentence) -> ExpressionSequence
Encoder: ExpressionSequence -> Expression

This transformation will strip out the information from the SimpleSentence. If we want to carry the information around, I think the Embedding struct that carry the real embedding and the pointer to the original input will make the information sharing less painful. For example, if we want to incroporate the segmentation information / other annotation of the input sentence, we can therefore annotate the SimpleSentence data structure and still use it in the encoder.

I think it is not very difficult to make the modular encoder API to still work, (so I will just rewrap the return value of the first encoder to the new Embedding object).

xnmt/translator.py
philip30: can you think of something more general than this :)

xnmt/input.py
philip30: so this isn't needed anymore?

This may reappear because of the auto merge, so shall I remove the whole class?

xnmt/encoder.py
philip30: thanks for pointing that out!

xnmt/input.py
philip30: done!

examples/data/flickr_tr20.ids
```diff
@@ -1,15 +0,0 @@
-0
```
neubig: Don't remove these. I think they're used in `speech_retriever.yaml`.

examples/flickr40k_retrieval.yaml
```diff
@@ -0,0 +1,49 @@
+defaults:
```
neubig: Remove this.

xnmt/translator.py
neubig: What about `calc_encoder_loss` or `calc_additional_loss`?

neubig:
Also, regarding how to carry around the input along with the encoding embeddings, is there a way to defer this until a later commit? If possible it'd be nice to discuss the best design separately.

philip30:
@neubig Thanks for the review. PTAL

neubig:
Thanks, looks good!

PR #155, opened 2017-08-05
modify fully-connected encoder

lwang114:
Some refactoring: use 'elif' instead of else: if:.


neubig:
It looks like this commit is changing much more than if->elif. Are all of the changes intentional?

lwang114:
Sorry I need to pull the latest version of the encoder.m

PR #156, opened 2017-08-06
Masking

neubig:
Currently, xnmt does not perform masking on the encoder inputs for each mini-batch. This can be problematic if batching is used at training time and not at test time, as it will lead to different results. This PR will create a framework for masking within each component of xnmt so that mini-batched results will be equal to non-minibatched results in most cases.


neubig:
@philip30 This PR includes a modification of the loss.

philip30:
It looks good and clean to me. Several comments though:
1. What is masking, btw? I am a bit lost here and do not understand why is it performed in training but not in testing.
2. In ExpressionSequence, if kwargs is not passed within the function, why don't just list all the parameter in the function header?

PR #158, opened 2017-08-08
Fixed on Segmentation + Report

philip30:
- Added a capability to print the segmentation report (html + file).
- Fixed the additional loss problem in segmenting_encoder.
- also, includes the bug fix for the report that only prints the last input.
- Added recursive sum that accumulates the call if result != None. It used in additional loss computation.


philip30:
Also added refactoring of HTMLReportable to Reportable. It is basically the same, but more generic one. So now we can factor the report to html_report() or file_report() and it is possible to extend to other kinds of reports with the same framework. The report will be written every time an example is decoded.

neubig:
Thanks, this is good. I also got a request to write a report in another format, and the number of formats may expand indefinitely. How about we just have a `report()` function that takes a string `report_format` as input that can be "html" "pickle" or whatever.

philip30:
I think all those reports can be classified as file_report(). In this file_report we can write pickle, writing to file or whatever, depending on the class's logic that implements that. I still want to separate the html_report (1 function that implements all the reports logic seems very dirty) with the others though.. If an extension is wished, one just can add one more function in the reports.py with recursive tag.

neubig:
OK, fair enough.

PR #160, opened 2017-08-09
Mingxing

mingxingdu:
Hello Graham,

I added padding before pass to conv layer for Harwath's encoder and Tilburg encoder

I tested both of them with speech_retrieval.yaml and flickr40k_retrieval.yaml

Thanks for your review


xnmt/specialized_encoders.py
```diff
@@ -8,6 +8,26 @@
...
 # This is a CNN-based encoder that was used in the following paper:
 #  http://papers.nips.cc/paper/6186-unsupervised-learning-of-spoken-language-with-visual-context.pdf
+
+def padding(src, src_height, src_width,filter_width, stride, batch_size, channel=1):
+   ''' do padding for the sequence input along the time step (for example speech), so that so that the output of convolutional layer has the same size(time) of the input.
+
+       note that for padding image(two dimensional padding), please refer to dyne.conv2d(..., is_valid = False)
+   '''
+   # make sure the src is the right shape 
+   src = dy.reshape(src, (src_height, src_width, channel), batch_size=batch_size)
```
neubig: Do you need the reshape here? This wastes memory, and I don't think it's not necessary.

mingxingdu:
Ok I will test without reshape.

xnmt/specialized_encoders.py
mingxingdu: I can remove it ! 

It's done in my push just now

neubig:
Thanks!

PR #161, opened 2017-08-10
Some Fixes

philip30:
Few fixes for:
- Preproc not working in python3
- matplotlib not working on a machine that does not run X-Server


neubig:
Does preproc work on Python 2 after this PR?

philip30:
Well, I guess so. I will write the test code later if wished.

neubig:
Yeah, let's write a test and try to run it on Python2. The testing code now
works, so it should be easy to write a new unit test.

Graham

philip30:
@neubig I added the test and also split the test into several files.

PR #162, opened 2017-08-10
Better packaging

philip30:
You can do:

```python3 setup.py develop```

To install python "live" package directory that changes over time.
To establish the PATH, we just need to ```import xnmt``` before importing any modules from xnmt. (This should be done in executables)


neubig:
Thanks! We should probably also update `requirements.txt`, but that can be a future PR.

PR #164, opened 2017-08-14
Weight noise in embeddings

msperber:
I added weight noise as a global option because I'm planning on adding it to LSTMs as well.


neubig:
I think this is OK for now, but we might have to re-think having a lot of things be in the global variables later.

msperber:
Yes I agree!

PR #166, opened 2017-08-15
remove unused package import

msperber:
I'm not sure what the purpose of this import was, but it's throwing me an error and doesn't seem to be used anywhere.

### DIFF
--- a/xnmt/xnmt_run_experiments.py
+++ b/xnmt/xnmt_run_experiments.py
@@ -14,7 +14,6 @@ import random
 import numpy as np
 
 # XNMT imports
-import xnmt
 import copy
 import model_globals
 import xnmt_preproc, xnmt_train, xnmt_decode, xnmt_evaluate


philip30:
Hey I added this. You should do ```python setup.py develop```
The import statement is used to append the package path to PYTHONPATH so xnmt modules can be called anywhere.

Please refer to this pull request:
https://github.com/neulab/xnmt/pull/162

neubig:
@philip30 Woops, sorry! I wonder what the conditions are for this working? If it's necessary but might cause confusion maybe we can comment it?

msperber:
I see, sorry! I guess I don't really understand what you mean by "xnmt modules can be called anywhere"?

philip30:
@neubig @msperber No problem, I should have commented that :)

What I meant by that was, for example, if we want to run the run_experiments from our working directory where the corpus in the yaml file is a relative path. We can run the $PATH_TO_XNMT/xnmt/xnmt_run_experiments from that directory. 

Currently, I believe that the only workaround is to run the script from the root of XNMT, right?
I don't know if it is achievable with different method other than adding the modules path to the PYTHONPATH. The import xnmt actually does that automatically. Also, I am not sure if this behavior is desirable.

msperber:
I see! I personally always used absolute paths in the config files because I wouldn't have expected it to work otherwise, so I hadn't encountered that problem. I'm not sure which way is better but we should probably at least make it easy for people to understand what's going on?

philip30:
OK, I will try to create Error message when importing failed.

PR #167, opened 2017-08-15
Test debug config

msperber:
To move a bit forward with automatic testing:
- fixed unit tests so they can be run by typing ```python -m unittest discover```
- commented out one test that didn't pass
- added an integration test that runs debug.yaml

I'm not sure what plans on CI are? In the meantime we should probably at least run these tests before creating a pull request.


neubig:
Thanks! I tried to turn travis before but I wasn't able to do it because my lab org hadn't been activated yet and it took time. I just turned it on now, so it'd be nice to get CI working.

PR #168, opened 2017-08-15
Input feeding + init_decoder

philip30:
Please double check this.
For initializing the decoder state:

encoding = encoder.encoder(embedder.embed(input_sent))
decoder.lstm.c = encoding[-1]
decoder.lstm.h = tanh(encoding[-1])


neubig:
Thanks! Did you run any quick tests to see if this affects accuracy?

philip30:
I haven't done it. I can try to run it on BTEC but it may take 1-2 days for the results.

philip30:
Do you mean arctan there? I am really not sure what's going on...

philip30:
Ah OK! I got it, nevermind. It was tanh in the decoder.

msperber:
Looks great & thanks for working on this!

I have a comment about the input feeding: I don't think we would need a linear projection here, probably the better solution would be concatenation and adjusting the LSTM's input dim accordingly? The LSTM has it's own linear projection of the inputs, so it would seem strange to have 2 in a row, unless I'm missing something.

Also I'm wondering about having multiple layers in the encoder and decoder. I saw OpenNMT has a copy option which requires equal number of layers in encoder and decoder and will initialize every decoder layer accordingly, and a dense option which will do a linear transform (didn't check the code, but I'm guessing from every encoder final state to every decoder initial state?). This would require a bit of restructuring of the encoders, so probably better to leave it for a future commit..

philip30:
Thanks for the comment @msperber. Right. I  also thought about that before, But were stuck when we want to do decoding at the first time step. What kind of vector should we add? Is it like zero vector?

msperber:
Hmm yeah, zero vector sounds reasonable to me..

neubig:
Thanks for the check @msperber ! @philip30 , I'll wait until you're done reflecting this to merge.

philip30:
@neubig The add_inputs return list of states where each state in the list consist of tuple of 2 states.
What are they? Is this the state for every item in the minibatch?

philip30:
@msperber @neubig PTAL

neubig:
The input feeding looks good, but the logic of decoder initialization is still the same as before, right? Are you planning on revising this? It might be a little bit more difficult than the input feeding part, so we could potentially save that for another commit.

philip30:
@neubig I think it should be done in the decoder (Because it is an initialization of the decoder, so it should be invisible from the encoder, IMO). But I changed the encoder part according to what you said

xnmt/encoder.py
```diff
@@ -36,17 +36,27 @@ def transduce(self, embed_sent):
     """
     raise NotImplementedError('Unimplemented transduce for class:', self.__class__.__name__)
 
+  def get_final_state(self):
+    """ Return the state that represents the transduced sequence """
+    return None
```
philip30: It means I need to implement this method to all the subclass of the Encoder. There are some encoders that are not used by the translator (only retriever). And get_final_state() is maybe confusing there. So instead I handle in the SoftmaxMLPDecoder, because it is used only in the translator and will throw an Error if the final_state is None.

xnmt/decoder.py
```diff
@@ -41,29 +45,67 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
...
+    self.h_t = None
 
-  def initialize(self):
-    self.state = self.fwd_lstm.initial_state()
+  def initialize(self, h_f):
+    if not h_f:
+      raise ValueError("The encoder does not support get_final_state() yet. " +
+                       "Error in initializing decoder.")
+    state = self.fwd_lstm.initial_state()
+    state = state.set_s([h_f, dy.tanh(h_f)])
```
philip30: Because h_f is the state of the LSTM in the encoder. I am passing this state as the first state of the LSTM in the decoder.

xnmt/encoder.py
neubig: Our current behavior is:
* `Retriever`: never call `get_final_state` so behavior is irrelevant.
* `Translator`: call `get_final_state` and implement special error handling that throws an error when we receive `None`.

What if we implement a similar class `Tagger` that also calls `get_final_state`. Then we'd have to implement the same error handling in `Tagger` as well. Thus, error handling should be done at the source of the error (here) instead.

xnmt/decoder.py
neubig: The correct way to pass the state from the encoder to the decoder (assuming the encoder and decoder are the same size is:
`decoder_state.set_s(encoder_state.s())`. So I think the function here should just be `set_s(h_f)`

Also, on a related note: could you document the `initialize` function here so we can tell what input `h_f` expects?

xnmt/encoder.py
```diff
@@ -36,17 +36,27 @@ def transduce(self, embed_sent):
...
     else:
       raise NotImplementedError("Unimplemented transduce logic for class:",
                                 self.builder.__class__.__name__)
-    return ExpressionSequence(expr_list=out, mask=embed_sent.mask)
+    output_expr = [dy.concatenate([state_f.h()[-1], state_b.h()[-1]]) for state_f, state_b in states]
+    self.final_encoder_state = states[-1]
+    return ExpressionSequence(expr_list=output_expr, mask=embed_sent.mask)
+
+  def get_final_state(self):
+    return dy.concatenate([self.final_encoder_state[0].s()[-1], \
```
neubig: I think this should be:

    return [dy.concatenate(f,w) for f, w in zip(self.final_encoder_state[0].s(), self.final_encoder_state[1].s())]

neubig: Also, if the decoder is not bidirectional will this fail?

xnmt/encoder.py
philip30: Fair enough, I will do so accordingly.

xnmt/encoder.py
philip30: Yes, I think this will fail. Good point. Bahdanau et al only used the backward state though. So what do you think we should do? Shall make a transformation to adjust the size of the hidden state?

neubig: I think in `get_final_state` we should just check the number of elements in final_encoder_state and act appropriately (e.g. just return `self.final_encoder_state[0].s()` in the case of unidirectional nets). Then (as @msperber noted), in `Translator` we can add several options for how to pass the state from encoder to decoder. The simple method we have now will fail in cases where the number of layers or number of elements in the encoder is different than the decoder, but that's fine for now.

philip30:
PTAL @neubig

neubig:
Looks good, thanks!

msperber:
Currently this breaks the test cases in cases the encoder/decoder don't match; we should probably make the decoder initialization optional? E.g. like in OpenNMT bridge: copy|none and we'll add other options later.

neubig:
I agree!

PR #171, opened 2017-08-17
word dropout

msperber:
- [X] implement word (type) dropout as in https://arxiv.org/abs/1512.05287
- [X] slightly unrelated bugfix: disable embedding weight noise at inference time
- [X] quick convergence test
- [X] simplified usage of recursive methods

### DIFF
--- a/examples/debug.yaml
+++ b/examples/debug.yaml
@@ -25,6 +25,7 @@ defaults:
     model: !DefaultTranslator
       src_embedder: !SimpleWordEmbedder
         emb_dim: 64
+        word_dropout: 0.1
       encoder: !LSTMEncoder
         layers: 1
         input_dim: 64
@@ -35,6 +36,7 @@ defaults:
         input_dim: 64
       trg_embedder: !SimpleWordEmbedder
         emb_dim: 64
+        word_dropout: 0.2
       decoder: !MlpSoftmaxDecoder
         layers: 1
         mlp_hidden_dim: 64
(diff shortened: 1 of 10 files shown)


msperber:
@philip30 I'm a bit confused for how I can make ```set_train()``` work in the embedders. Maybe you could help me out? I called ```register_hier_child()``` [here](https://github.com/neulab/xnmt/pull/171/files#diff-fd01cb9452fd023aa34552ca193efb5cR74). The problem I have:
- If I have ```SimpleWordEmbedder``` inherit from ```model.HierarchicalModel```, I get this error: 'SimpleWordEmbedder' object has no attribute '_hier_children'
- I I don't, I get a warning and set_train() is never called: ('Skipping hierarchical construction:', 'SimpleWordEmbedder', 'is not a subclass of HierarchicalModel')

philip30:
Hi @msperber, I think it is because you didn't call the super constructor in the SimpleWordEmbedder. The attribute is constructed there. I think this is not a good behaviour and confusing... Maybe you can change it a bit so it will be an attribute even though we are not calling the constructor of the HierarchicalModel?

msperber:
Thanks, I got it! I refactored things a bit so we don't have to call the super constructor, hope it's okay. From my side this would be ready for merging then.

neubig:
LGTM if @philip30 is OK.

philip30:
LGTM

PR #172, opened 2017-08-21
moved some configs from examples to tests

msperber:
I guess we should aim at putting config files for purposes of documentation and integration testing in test/config/ and config files with reasonable settings that are supposed to produce good results in examples/

3 config files (segmenting.yaml, segmenting2.yaml, retrieval.yaml) are broken and need to be fixed.


neubig:
Thanks! I agree with this general policy, but for some of these it's not entirely clear (e.g. `report.yaml` is also an example of how to use the reporting functionality). For now this looks good though, so you can merge after resolving the conflicts.

PR #173, opened 2017-08-22
LSTM init

msperber:
Just thought I'd open this to let people know I'm working on this.

This will eventually:
- make decoder initialization optional
- allow copying of final states from arbitrary encoders, including non-LSTM-based encoders
- allow multiple encoder / decoder layers for copying, as long as encoder-layers >= decoder_layers
- Add Python LSTM code for more convenience / independence from the Dynet C++ implementation


msperber:
This should be ready to merge, could you take a look @neubig @philip30 ? FYI I removed ConvBiRNNBuilder because it's not a very general encoder and not worth maintaining.

philip30:
It looks totally good for me. One comment: shall we set the default brigde to NoBridge? This push will break the existing translation YAML file I think.

msperber:
Yeah it would be nice to set a default! Unfortunately I think it's not possible currently to set an instantiated object as default due to how the YAML deserialization / parameter sharing works, but would be a nice thing to have in the future.

philip30:
How about setting it to NoBridge in the constructor if the bridge if None? This will preserve the backward compatibility I think.

neubig:
I'm guessing it's not that simple according to @msperber 's explanation, but if it is that simple I'm in favor of doing it. If not, I think this PR is OK to merge for now, and we can think of a way to fix it later.

msperber:
None is not possible because currently parameter sharing will crash when it encounters a None object, so I'll merge this for now!

PR #175, opened 2017-08-24
Tests, improvements, fixes

msperber:
- Fixed dynet imports that went broken after recent dynet update
- Added a few test cases, improved code for better modularity/testability/readability along the way
  - among other changes, start_sent() is now a recursive method and can be implemented anywhere in the hierarchical model

One detail that was different between training / decoding was that for decoding we were feeding in the beginning-of-sequence token as first input to the decoder LSTM, while at training time we weren't doing that. Not sure which way is the correct way? Currently I changed the decoder to match the training behavior..


msperber:
I've added additional tests + fixes, but this is about as far as I was planning to write tests so it could be merged if there are no concerns.

test/test_beam_search.py
```diff
@@ -0,0 +1,156 @@
+import unittest
+
+import _dynet as dy ; dyparams = dy.DynetParams() ; dyparams.set_random_seed(13); dyparams.init()
```
neubig: This is not the preferred way of doing this anymore. Could you please try the new way?
http://dynet.readthedocs.io/en/latest/python_ref.html#dynet.DynetParams

msperber: I updated it everywhere except here, will that be alright?
https://github.com/neulab/xnmt/pull/175/files#diff-d11c4a7987fb2f1114fff16beb26c6ffR163

xnmt/decorators.py
```diff
@@ -8,7 +6,8 @@ def recursive(f):
   '''
   # Wrapper to the method to call recursively, instead of invoking only f()
   def rec_f(obj, *args, **kwargs):
-    assert(issubclass(obj.__class__, model.HierarchicalModel))
+    from model import HierarchicalModel
```
philip30: Why are we importing classes in the middle of the code? Does this mean that this import will be called over and over again?

test/test_beam_search.py
msperber: Hm, I got a duplicate initialization warning so I removed dyparams.init() in xnmt_run_experimnts.py.
Hope this is the correct way to do it:
```
import _dynet
dyparams = _dynet.DynetParams()
dyparams.from_args()
```

xnmt/decorators.py
msperber: I did this to fix problems with a circular import: previously, model.py imported decorators.py, and decorators.py imported model.py.

Python will only execute the first import, so it should be okay, but would probably be even better to improve the structure so we don't need to worry about this?

philip30: right. I think the better structure will be putting the HierarchicalModel and all of the decorators together in the same file (hier_model.py). It is up to you though if you want to change it that way, or can be done in a separate pull request.

msperber: Sure, let's do it one a later PR then! Part of this PR is uncovering a few structural issues, for example this one or also some things in xnmt_train.py which I noticed while writing tests but won't be able to fix myself right now.

neubig:
Are unit tests passing for you? In my environment with the latest master version of DyNet I get the attached errors.
[err.txt](https://github.com/neulab/xnmt/files/1260538/err.txt)

philip30:
BTW, in the encoder.py line 132, will break Python3. The reduce keyword is removed in python3. This is quoted from stack overflow:

```
Removed reduce(). Use functools.reduce() if you really need it; however, 99 percent of the time an explicit for loop is more readable.
```

msperber:
Thanks for the hints! I'll merge since unit tests are passing now (except for one that requires a fix by Philipp)

philip30:
@msperber Which one is that? All unittests passed in my environment.

msperber:
Me too now, I guess you already fixed it ;-)

philip30:
@msperber Hey, I have a quick question. So now every Serializable has a context in their constructor? Is it necessary?

msperber:
Not every Serializable, only the ones that need it. The YAML serializer is automatically passing this if needed, so that we can get rid of the module-global model_globals which caused all kinds of trouble.

philip30:
@msperber That's cool. How to determine if one object needs a context?

msperber:
By using Python inspect, it checks if the class has "context" as one of its constructor parameters!

PR #180, opened 2017-09-09
Resolving inconsistencies in Beam Search

psjanani:
Fixing Beam Search to be compatible with the current Decoder.

The following discrepancy was observed by @msperber when using Beam Search for inferencing using a trained model:
```
beam	BLEU	
1	0.294	
2	0.269	
3	0.269	
4	0.265
```	

Test:
Testing on the above model now provides consistency in inference scores. And following BLEU scores were observed:
```
beam	BLEU4	
1	0.2652	
2	0.2751	
3	0.2794
```

Tested on the unit-test suite

### DIFF
--- a/xnmt/decoder.py
+++ b/xnmt/decoder.py
@@ -81,6 +81,12 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
     return [set(["layers", "bridge.dec_layers"]),
             set(["lstm_dim", "bridge.dec_dim"])]
 
+  def get_state(self):
+    return (self.state, self.h_t)
+
+  def set_state(self, params):
+    self.state, self.h_t = params
+
   def initialize(self, enc_final_states):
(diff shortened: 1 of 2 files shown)


neubig:
Thanks a lot for looking at this! LGTM, but hopefully @msperber can confirm that this works and hopefully figure out why the BLEU scores are different across systems.

msperber:
Thank you, the changes also look good to me as far as I can tell. But I also observed the drop for beam size 1 from 29 to 26 so there must still be something wrong. Could you confirm if you get around BLEU 29 for beam size 1 before this fix?

psjanani:
Hi! The values before the fix are the following for me:
```
beam	BLEU4	
1	0.2652	
2	0.2457	
3	0.2433
```
The value is consistent for beam=1. But, I am unable to reproduce the results obtained by @msperber.

PR #181, opened 2017-09-12
Segmenting encoder

philip30:
I would like to merge the current stable segmenting_encoder branch to master.
It includes some fixes to the current SegmentingEncoder in master.


neubig:
Looks good! I haven't checked segmenting_encoder.py, but I think that's under active development. The other parts look OK.

PR #184, opened 2017-09-20
transformer encoder

DevSinghSachan:
Hi Graham, Matthias,
I have added code for the 1 layer transformer encoder from Attention is all you need paper. This includes multi-head attention and Position feed forward layer.
Currently, masking is not done.

Thanks,

### DIFF
--- a/examples/transformer.yaml
+++ b/examples/transformer.yaml
@@ -32,10 +32,9 @@ defaults:
         input_dim: 512
       trg_embedder: !SimpleWordEmbedder
         emb_dim: 512
-      decoder: !MlpSoftmaxDecoder
+      decoder: !SoftmaxDecoder
         layers: 1
-        mlp_hidden_dim: 512
-        bridge: !CopyBridge {}
+        bridge: '!NoBridge'
   decode:
     src_file: examples/data/test.ja
   evaluate:
--- a/examples/transformer.yaml
+++ b/examples/transformer.yaml
@@ -1,0 +1,46 @@
+defaults:
+  experiment:
+    model_file: examples/output/<EXP>.mod
+    hyp_file: examples/output/<EXP>.hyp
+    out_file: examples/output/<EXP>.out
+    err_file: examples/output/<EXP>.err
+    run_for_epochs: 20
+    eval_metrics: bleu
+  train:
+    default_layer_dim: 512
+    restart_trainer: True
+    trainer: Adam
+    learning_rate: 0.0002
+    lr_decay: 0.5
+    dev_metrics: bleu
+    training_corpus: !BilingualTrainingCorpus
+      train_src: examples/data/train.ja
+      train_trg: examples/data/train.en
+      dev_src: examples/data/dev.ja
+      dev_trg: examples/data/dev.en
+    corpus_parser: !BilingualCorpusParser
+      src_reader: !PlainTextReader {}
+      trg_reader: !PlainTextReader {}
+    model: !DefaultTranslator
+      src_embedder: !SimpleWordEmbedder
+        emb_dim: 512
+      encoder: !TransformerEncoder
+        layers: 1
+      attender: !StandardAttender
+        hidden_dim: 512
+        state_dim: 512
+        input_dim: 512
+      trg_embedder: !SimpleWordEmbedder
+        emb_dim: 512
+      decoder: !MlpSoftmaxDecoder
+        layers: 1
+        mlp_hidden_dim: 512
+        bridge: !CopyBridge {}
+  decode:
+    src_file: examples/data/test.ja
+  evaluate:
+    ref_file: examples/data/test.en
+
+standard-dropout0.1:
+  train:
+    dropout: 0.1
--- a/examples/transformer.yaml
+++ b/examples/transformer.yaml
@@ -9,9 +9,10 @@ defaults:
   train:
     default_layer_dim: 512
     restart_trainer: True
-    trainer: Adam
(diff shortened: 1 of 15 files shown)


msperber:
Sounds nice! I have a minor comment regarding code structure: Should we maybe move linear.Linear, mlp.MLP, transformer.AffineTransform, transformer.PositionwiseFeedForward to a general purpose "nn.py" module?
Also, the default dropout for the TransformerEncoder should probably be 0.0, to be consistent with the other components.

I haven't checked the network code in detail but hope to play around with it myself some when I find some time (not this week).

DevSinghSachan:
Hi Matthias,
Thanks for the comments !
I have refactored the code into nn.py module and set the initiated the dropout value to 0.

I was thinking that in order to implement decoder module, we can do it with some changes to the attention Class. And that we can use directly compute the losses in the decode step for the entire sentence length using appropriate target masks. (suggested by @neubig )

So, I was thinking that should we write a new translation Class , something like TransformerTranslator() etc. and make changes the calc_loss() function.

neubig:
Thanks @DevSinghSachan ! I'll try to take a look at this.

One thing I'd like you to try to be careful of is making a *minimum* of specialized functions. Any functions that are generalized and reusable, try to make them so. The best thing possible would be the ability to create `transformer.yaml` that simply used other reusable components.

That being said, I know this is not always possible, or not possible to do easily. In this case, if you need something like `TransformerTranslator`, let's add it to something like `specialized_translators.py` (similar to `specialized_encoder.py`). We can always modularize it later.

philip30:
Hey @DevSinghSachan  this is minor, but can you make sure that you are using 2 tab space? I think it is not good to mix them up :)

xnmt/encoder.py
```diff
@@ -311,6 +321,24 @@ def transduce(self, embed_sent):
   def initial_state(self):
     return PseudoState(self)
 
+
+class TransformerEncoder(BuilderEncoder, Serializable):
+  yaml_tag = u'!TransformerEncoder'
+
+  def __init__(self, context, input_dim=512, layers=1, hidden_dim=512, dropout=0.5, **kwargs):
```
msperber: Sorry, actually what I meant regarding dropout would be something like:

def __init__(self, context, input_dim=512, layers=1, hidden_dim=512, dropout=None, **kwargs):
  self.dropout = dropout or context.dropout

That way we will be using the globally configured dropout, unless the TransformerEncoder-specific dropout value is set in the config.

examples/transformer.yaml
```diff
@@ -0,0 +1,46 @@
+defaults:
```
msperber: One more suggestion from my side would be to extend test/config/encoders.yaml to include a model with the new encoder (similar to this config but maybe with a tiny hidden size). Just to make sure future configs won't break this encoder.

DevSinghSachan: That's a good idea. I will do this. thanks for the suggestions !!

test/test_beam_search.py
```diff
@@ -6,7 +6,7 @@
 from xnmt.translator import DefaultTranslator
 from xnmt.embedder import SimpleWordEmbedder
 from xnmt.encoder import LSTMEncoder
-from xnmt.attender import StandardAttender
```
neubig: Fix to revert to StandarAttender

DevSinghSachan:
Hi Graham and Matthias,

Transformer Encoder and Decoder along with masks support is completed.

Here are list of things to be done-
a)  Fix for label smoothing (code in separate branch)
b)  Modify `calc_loss` function in translator / decoder to use the output from TransformerDecoder.
c)  Write `generate` function for TransformerTranslator class.
d)  Write a class for special training procedure for Adam Optimizer as mentioned in 'AIAYN' paper.

neubig:
Thanks! Have you tested to see if this gets reasonable accuracy?

DevSinghSachan:
Hi Graham,

I have only done initial testing for the word train_loss/word. After 6 epochs using standard configurations, I am able to get a train_loss/word of 0.962 .
Currently, I working on getting BLEU score metric, in order to test the code.

neubig:
OK, cool! That will be an important indicator of whether the code is correct or not, so I'll wait until we have an accuracy number before merging.

msperber:
BTW, parts of the code could probably be simplified using the refactored masking code #189 , in case you want to make use of that.

DevSinghSachan:
Hi Matt,
Thanks a lot for the suggestion!
I will look into the integration with refactored mask code once I initially get the Transformer model working.

DevSinghSachan:
Currently, the performance of the model is not at pr with the standard example.
I am attaching the logs here.

Here is the log of the using a 1 layer model on the standard data set of ja->en
[TF.log](https://github.com/neulab/xnmt/files/1352246/TF.log)

Here is a sample of generated sentences.
[sent_gen.txt](https://github.com/neulab/xnmt/files/1352252/sent_gen.txt)

xnmt/expression_sequence.py
```diff
@@ -32,7 +34,7 @@ def __len__(self):
     :returns: length of sequence
     """
     if self.expr_list: return len(self.expr_list)
-    else: return self.expr_tensor.dim()[0][0]
+    else: return self.expr_tensor.dim()[0][1]  # Earlier it was [0][0], but 0 is the number of dims
```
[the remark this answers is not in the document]
msperber: Sorry nevermind, you are correct.

PR #185, opened 2017-09-21
Reinforce loss with cythonize BLEU

philip30:
First you need to compile the cython code by:

```python setup.py build_ext --inplace```


neubig:
What happens when you try to run xnmt without running the cython step? I'd like it to be optional if possible, as this is an uncommon use case and it may be a barrier to installing the toolkit.

philip30:
In that case, I have 2 options:
- Push the binary and the generated CPP code to the repository.
- Try catch the code where it tries to use cython and falls to the python implementation if no such code exists. (Will be like try catch in the import)

However, I recommend to still use the cython. Every time this kind of error happens, we can just run the ```python setup.py develop``` again to fix it (the setup.py will compile the code.

philip30:
I move the import to before the actual code use. This will prevent the error you mention. But it still won't pass the test case.

neubig:
Sorry, my comment was not clear. Basically my concern is that cython is one more moving part that could break in some user's environments. If this is an absolute requirement for using xnmt, every cython breaks we'll have to go in and debug why it happened and make a fix. This is annoying (I have to do this a lot for DyNet).

Basically, if we're using Cython for only a tiny subset of the functionality in xnmt, I'd like the Cython step to be optional, so we only need to debug it for power users who are using this functionality. If someone who has not run the cython step tries to use cython-dependent functionality, then we can throw an error "this functionality requires compilation of xnmt's C++ code using cython. Please see the Cython installation instructions for more details".

philip30:
What do you mean by "cython step to be optional?" Does that mean that it should not be contained in the normal setup.py?

neubig:
It can be included in setup.py, but be an option. For example, if we add the "--use-cython-extensions" option, then it will use the cython extensions.

philip30:
Done!

philip30:
@neubig Is this ready to merge?

neubig:
Sorry for the delay!

PR #186, opened 2017-09-24
added label smoothing term in cost function

DevSinghSachan:
added additional term for Label Smoothing in cost function for model regularisation.
This has been shown in a couple of papers to give better BLEU scores and dev loss as it acts to regularise the model.

From my experiments, on the big dataset of jp->en translation task, I got the following results on dev set
i) BLEU: 26.211 (    0.1\*label smoothing   + 0.9\*CE  loss)
ii) BLEU: 25.433  ( only CE  loss )

### DIFF
--- a/examples/standard.yaml
+++ b/examples/standard.yaml
@@ -36,6 +36,7 @@ defaults:
         layers: 1
         mlp_hidden_dim: 512
         bridge: !CopyBridge {}
+        eps: 0.1
   decode:
     src_file: examples/data/test.ja
   evaluate:
@@ -44,3 +45,4 @@ defaults:
 standard-dropout0.1:
   train:
     dropout: 0.1
+
--- a/examples/standard.yaml
+++ b/examples/standard.yaml
@@ -36,7 +36,7 @@ defaults:
         layers: 1
         mlp_hidden_dim: 512
         bridge: !CopyBridge {}
-        eps: 0.1
+        label_smoothing: 0.1
   decode:
     src_file: examples/data/test.ja
   evaluate:
--- a/xnmt/decoder.py
+++ b/xnmt/decoder.py
@@ -1,3 +1,5 @@
+from __future__ import division
+
 import dynet as dy
 from xnmt.serializer import Serializable
 import xnmt.batcher
@@ -40,13 +42,15 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
   def __init__(self, context, vocab_size, layers=1, input_dim=None, lstm_dim=None,
                mlp_hidden_dim=None, trg_embed_dim=None, dropout=None,
                rnn_spec="lstm", residual_to_output=False, input_feeding=False,
-               bridge=None):
+               bridge=None, eps=0.1):
     param_col = context.dynet_param_collection.param_col
     # Define dim
     lstm_dim       = lstm_dim or context.default_layer_dim
     mlp_hidden_dim = mlp_hidden_dim or context.default_layer_dim
     trg_embed_dim  = trg_embed_dim or context.default_layer_dim
     input_dim      = input_dim or context.default_layer_dim
+    self.vocab_size = vocab_size
+    self.eps = eps
     # Input feeding
     self.input_feeding = input_feeding
     self.lstm_dim = lstm_dim
@@ -116,7 +120,9 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
       return dy.pickneglogsoftmax(scores, ref_action)
     # minibatch mode
     else:
-      return dy.pickneglogsoftmax_batch(scores, ref_action)
+      ce_loss_term = (1 - self.eps) * (dy.exp(-dy.pickneglogsoftmax_batch(scores, ref_action)))
+      label_smoothing_term = self.eps / self.vocab_size
+      return -dy.log(ce_loss_term + label_smoothing_term)
 
   @recursive
   def set_train(self, val):
--- a/xnmt/decoder.py
+++ b/xnmt/decoder.py
@@ -115,6 +115,11 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
     return self.vocab_projector(self.h_t)
 
   def calc_loss(self, context, ref_action):
+    """
+    Label Smoothing is implemented with reference to Section 7 of the paper
+    "Rethinking the Inception Architecture for Computer Vision"
+    (https://arxiv.org/pdf/1512.00567.pdf)
+    """
     scores = self.get_scores(context)
     # single mode
     if not xnmt.batcher.is_batched(ref_action):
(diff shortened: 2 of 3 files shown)


examples/standard.yaml
```diff
@@ -36,6 +36,7 @@ defaults:
         layers: 1
         mlp_hidden_dim: 512
         bridge: !CopyBridge {}
+        eps: 0.1
```
neubig: Make this a more descriptive name, e.g. "label_smoothing".

xnmt/decoder.py
```diff
@@ -116,7 +120,9 @@ def calc_loss(self, context, ref_action):
       return dy.pickneglogsoftmax(scores, ref_action)
```
neubig: This needs to be implemented in non-minibatch mode as well.

xnmt/decoder.py
```diff
@@ -116,7 +120,9 @@ def calc_loss(self, context, ref_action):
...
     # minibatch mode
     else:
-      return dy.pickneglogsoftmax_batch(scores, ref_action)
+      ce_loss_term = (1 - self.eps) * (dy.exp(-dy.pickneglogsoftmax_batch(scores, ref_action)))
```
neubig: This will be much slower than pickneglogsoftmax, so when label smoothing is zero, we should use the normal pickneglogsoftmax.

examples/standard.yaml
DevSinghSachan: done via the latest commit !

xnmt/decoder.py
DevSinghSachan: done via the latest commit !

xnmt/decoder.py
DevSinghSachan: done via the latest commit !

xnmt/decoder.py
```diff
@@ -40,13 +42,16 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
   def __init__(self, context, vocab_size, layers=1, input_dim=None, lstm_dim=None,
                mlp_hidden_dim=None, trg_embed_dim=None, dropout=None,
                rnn_spec="lstm", residual_to_output=False, input_feeding=False,
-               bridge=None):
+               bridge=None, label_smoothing=0.1):
```
neubig: Should be 0.0

xnmt/decoder.py
```diff
@@ -113,10 +118,16 @@ def calc_loss(self, context, ref_action):
...
+      log_loss = dy.pickneglogsoftmax(scores, ref_action)
+
+    else: # minibatch mode
+      log_loss = dy.pickneglogsoftmax_batch(scores, ref_action)
+
+    if self.label_smoothing == 0.0:
+      return log_loss
     else:
-      return dy.pickneglogsoftmax_batch(scores, ref_action)
+      ce_term = (1 - self.label_smoothing) * (dy.exp(-log_loss))
```
neubig: Take another look at this and make sure it matches section 7 of https://arxiv.org/pdf/1512.00567.pdf

xnmt/decoder.py
```diff
@@ -40,13 +42,16 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
...
                rnn_spec="lstm", residual_to_output=False, input_feeding=False,
-               bridge=None):
+               bridge=None, label_smoothing=0.1):
     param_col = context.dynet_param_collection.param_col
     # Define dim
     lstm_dim       = lstm_dim or context.default_layer_dim
     mlp_hidden_dim = mlp_hidden_dim or context.default_layer_dim
     trg_embed_dim  = trg_embed_dim or context.default_layer_dim
     input_dim      = input_dim or context.default_layer_dim
+    self.label_smoothing = label_smoothing
```
neubig: Add reference to https://arxiv.org/pdf/1512.00567.pdf

xnmt/xnmt_decode.py
```diff
@@ -32,6 +32,7 @@
   Option("beam", int, default_value=1),
   Option("max_len", int, default_value=100),
   Option("len_norm_type", str, required=False),
+  Option("label_smoothing", float, required=True, default_value=0.1, help_str="label smoothing parameter")
```
neubig: Maybe set this to be 0.0 by default

xnmt/decoder.py
```diff
@@ -113,10 +118,16 @@ def calc_loss(self, context, ref_action):
...
+
+    else: # minibatch mode
+      log_loss = dy.pickneglogsoftmax_batch(scores, ref_action)
+
+    if self.label_smoothing == 0.0:
+      return log_loss
     else:
-      return dy.pickneglogsoftmax_batch(scores, ref_action)
+      ce_term = (1 - self.label_smoothing) * (dy.exp(-log_loss))
+      return -dy.log(ce_term + self.label_smoothing_term)
```
pmichel31415: Hi. I've implemented label smoothing in my own code and I think the implementation should be like:

```python
ce_term = dy.pick_batch(log_loss, ref_action)
label_smoothing_term = dy.mean_elems(log_loss)
final_loss = (1 - self.label_smoothing) * ce_term + self.label_smoothing * label_smoothing_term
```

Basically you are just changing the target distribution so you don't need to leave log-space to compute the new cross entropy. Also in this case mean_elems is probably faster than dividing by the vocabulary size.

I think what you're doing now is replacing  H(q, p) by H(q, (1- eps)p + eps u) instead of H((1-eps)q + eps u, p) (using the notations of https://arxiv.org/pdf/1512.00567.pdf

DevSinghSachan: Hi Paul,
Thanks a lot for the helpful comments. I have now re-implemented the label smoothing functionality as you have suggested.

I observe that the speed: words/sec have dropped by almost 30-40% when using label smoothing (mayne due to computing the softmax over all the words in vocab). Did you also observe something similar.

pmichel31415: I got something around a 5% slow-down... Maybe this is because you seem to be computing the softmax twice (in pickneglogsoftmax for the normal loss term and in log_softmax for the label smoothing term). 

Whenever I used label smoothing I would compute `log_prob=log_softmax(score)` and then do all operation on this expression (ie not doing `pickneglogsoftmax`, but `pick` on `log_prob`).

neubig:
@DevSinghSachan Is this fixed now? I'll be happy to check one more time once you fix the merge conflicts.

DevSinghSachan:
Thanks @neubig . The code is overall fixed. I will resolve the merge conflicts and send a PR with some timing info (wps) for comparing the speed of execution as well.

neubig:
Thanks! And no need to send a new PR. You can just update this one.

DevSinghSachan:
I have also implemented label smoothing here: https://github.com/DevSinghSachan/aiayn

I have just sent a message regarding this as well. I am thinking that, once everything seems fine in the above code, I will update both this branch and the transformer branch with the updated code.

xnmt/xnmt_decode.py
neubig: This is not fixed yet.

DevSinghSachan:
Opened a new pull request regarding this here: https://github.com/neulab/xnmt/pull/239

PR #188, opened 2017-09-28
Fix enc masking

msperber:
This can make a huge difference especially for "trg" batching strategy. Most encoder code now uses expression sequences because that seems the cleanest way to keep expressions and masks together.


neubig:
Cool, thanks!

PR #191, opened 2017-10-02
Prior Segmentation

philip30:
This pull request enables the segmenting encoder to be used with the prior segmentation data.

The config to run the SegmentingEncoder can be seen in the test/config/prior-segmenting.yaml
The example of the data can be seen in:

- examples/data/head.ja -> segmented data
- examples/data/head-char.ja -> character of segmented data
- examples/data/head-seg.ja -> the segmentation

Even though this is a translation, we can think the characters as speech frames, and thus the rest will be the same. I include @mingxingdu so he can be aware of it also.


philip30:
I don't know why the commits strange. Please see the file changes instead.

msperber:
Hey Philipp, just one comment from my side, as you already mentioned adding encoder.set_input() may not be ideal. I’m just wondering if it would be better to extend the @recursive start_sent() to -> start_sent(src)? That way we could avoid adding more complexities to the code.

neubig:
@philip30 Thanks! I think that we need to think about this semi-carefully. In the future, we might want other types of inputs as well (e.g. a tree-structured input for tree-based MT), so coming up with a general solution is probably a good idea. Perhaps we can talk about this a little bit at the lab meeting today.

philip30:
I think I have fixed the segmenting encoder prior segmentation with the current SeqTransducer model. I am merging it if there is no problem.

neubig:
Wait a second, did we ever come up with a solution for how we were going to properly take inputs with annotations such as the segment boundaries or parse trees? I don't think we did...

philip30:
At the moment, I just simply create a separate input class SentenceInput which we can annotate with anything. Then I create a separate Reader to annotate this SentenceInput. 

The regular SentenceInput without annotation is renamed into SimpleSentenceInput. But I think we should create separate PR to manage this input reading more properly. The current implementation is very easy to factor though. @neubig

neubig:
OK. On a big commit like this that includes design decisions it'd be nice to get someone's OK before merging though. (I've done several self merges, which is probably bad practice... I'll try to also get review before I merge.)

philip30:
@neubig We can still revert the pull request, if this is a problem. Sorry for that, let's discuss the design before merging this pull request.

PR #194, opened 2017-10-10
Made the decoder stateless

neubig:
This commit makes the decoder stateless, removing any of its global variables and instead holding them in a separate state object. This will probably reduce potential for bugs in search algorithms, etc.

Also, input feeding is put on by default. It also makes a change to input feeding, feeding in the context, not the hidden layer of the MLP. It might be worth feeding in the hidden layer of the MLP instead, but I don't think this is the standard way of doing things (as noted by @cindyxinyiwang )


neubig:
Woops, I did this with an old version of xnmt. Let me trash this pull request and retry.

PR #197, opened 2017-10-10
Pretrained embeddings loader

MattX:
Standard embedder that loads initial embeddings from a fasttext-format file.

The embedding dimension has to be specified in the config file, but (obviously) it has to be the same as in the fasttext file. This is a little redundant, but I couldn't figure out a nice way to support autodetection from the file and still automatically pass the embedding size to the encoder. I'm a little confused by some of the serialization code, so I may be missing something.

I also added a method to Tee that my IDE expects to be present in `sys.stdout` for unit test integration. This is completely unrelated to the rest, and I can open a separate pull request if needed, but since it's just two lines I though it'd be OK here.


xnmt/embedder.py
```diff
@@ -143,3 +144,72 @@ def embed_sent(self, sent):
           embeddings.append(self.embed(xnmt.batcher.mark_as_batch([single_sent[word_i] for single_sent in sent])))
       return ExpressionSequence(expr_list=embeddings, mask=sent.mask)
 
+
+class PretrainedSimpleWordEmbedder(SimpleWordEmbedder):
```
neubig: I think it'd be better to just add an option to let it read in embeddings from a file if the argument is not None, rather than creating another class. What do you think?

MattX: I agree; the only argument I could see for keeping a separate class is if we want to add some different logic to prevent pretrained weights from being updated / being changed too much.

neubig: Ahh, that's a good point. Actually now that you mention it, we might want to keep this as-is so we can add this functionality later.

PR #198, opened 2017-10-10
[WIP] Fixed some test problems

MattX:
- Fixed a few python2/3 compatibility problems
- Replaced enum with enum34, which hopefully should fix the Travis problem.
- I've observed that `test_translator_loss` and `test_overfitting` occasionally fail, so I've set the seed to 0 to make them reproducible. It may be worth investigating, but it's not very useful to have it make CI tests fail randomly.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ build/
 xnmt/cython/*.cpp
 xnmt/cython/*.c
 xnmt/cython/*.so
+test/tmp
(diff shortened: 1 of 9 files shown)


neubig:
Actually, I think a random seed of zero will still change every time. Could you try a non-zero number?

MattX:
Thanks for the tip, it looks like changing the seed fixed the problem. I think the builds are still failing because other tests have broken on master.

neubig:
Yay, green tests! Thanks a bunch @MattX

PR #207, opened 2017-10-13
component sharing

msperber:
This drafts some code to make it possible to share YAML components. The idea is to use YAML anchors / aliases like this:
`    model: !DefaultTranslator`
`      src_embedder: !SimpleWordEmbedder &src_emb`
`      encoder: ...`
`      attender: ...`
`      trg_embedder: *src_emb`
`      decoder: ...`
In the deserialization code, we need to catch if something is an alias, assume that the corresponding anchor has already been initialized (so proper ordering does matter), and then simple reference to that object instead of initializing a new one.

Problem is that pyyaml completely resolves anchors and aliases and doesn't return any info on this. Possibly something like this could help (don't have time to look into it right now): https://github.com/ChloeTigre/pyyaml-keep-anchors

Otherwise, we could make our own referencing mechanism, but according to my first attempt in that direction, that might introduce complications to the param sharing code..


msperber:
Actually I had a very simple idea for how this should work without installing additional packages etc, which I quickly wrote up.
It's untested but I expect the deserialization part of the sharing feature should basically work.

msperber:
This should be ready to merge: Component sharing works, and I included an example for (1) sharing reference to the same object in memory, (2) copying over component config, but creating separate objects, and (3) referencing individual values.

One caveat: the sharing of embedders from the example is actually not working properly, because the vocabs are not shared. To make this work properly, we would either have to move the readers inside the model, or separate vocab and readers and move at least the vocabs inside the model (I think that would be an intuitive thing to do anyways).

neubig:
Thanks, this looks really nice. One small comment: perhaps you could add a little bit of documentation to the beginning of the example yaml file? The types of sharing that you can do and syntax for realizing each of these was pretty difficult for me to grok based on what we currently have.

msperber:
Sure, how about this?

neubig:
Awesome, thanks! When you resolve the conflicts I think this is ready to merge.

PR #208, opened 2017-10-13
[WIP] Adding Baseline to Reinforce

psjanani:
Currently, the Regressor is trained at sentence level as the process is very slow otherwise.

### DIFF
--- a/xnmt/translator.py
+++ b/xnmt/translator.py
@@ -10,15 +10,18 @@ import xnmt.length_normalization
 import xnmt.batcher
 
 from xnmt.vocab import Vocab
+from xnmt.model import HierarchicalModel
 from xnmt.serializer import Serializable, DependentInitParam
 from xnmt.search_strategy import BeamSearch, GreedySearch
 from xnmt.output import TextOutput
+import xnmt.linear as linear
 from xnmt.model import GeneratorModel
 from xnmt.reports import Reportable
-from xnmt.decorators import recursive_assign, recursive
+from xnmt.decorators import recursive_assign, recursive, recursive_sum
 import xnmt.serializer
 import xnmt.evaluator
 from xnmt.batcher import mark_as_batch, is_batched
+from xnmt.loss import LossBuilder
 
 # Reporting purposes
 from lxml import etree
@@ -75,7 +78,8 @@ class DefaultTranslator(Translator, Serializable, Reportable):
       self.loss_calculator = TranslatorMLELoss()
     else:
       self.loss_calculator = loss_calculator
-
+    if (issubclass(self.loss_calculator.__class__, HierarchicalModel)):
+      self.register_hier_child(self.loss_calculator)
     self.register_hier_child(self.encoder)
     self.register_hier_child(self.decoder)
     self.register_hier_child(self.src_embedder)
@@ -234,36 +238,17 @@ class TranslatorMLELoss(Serializable):
 
     return dy.esum(losses)
 
-class TranslatorReinforceLoss(Serializable):
+class TranslatorReinforceLoss(Serializable, HierarchicalModel):
   yaml_tag = '!TranslatorReinforceLoss'
 
-  class Regressor():
-    def __init__(self, dim):
-      self.param_collection = dy.Model()
-      self.trainer = dy.AdamTrainer(self.param_collection)
-      self.pW = self.param_collection.add_parameters((1, dim))
-      self.pb = self.param_collection.add_parameters(1)
-
-    def predict_score(self, h_t, dim):
-      h_t = dy.inputTensor(np.reshape(h_t, (dim[0][0], dim[1])), batched=True)
-      W = dy.parameter(self.pW)
-      B = dy.parameter(self.pb)
-      output = B + W * h_t
-      return output
-
-    def calc_loss(self, h_t, true, dim):
-      output = self.predict_score(h_t, dim)
-      true = dy.inputTensor(true, batched=True)
-      err = dy.squared_distance(true, output)
-      return dy.sum_batches(err)
-
-  def __init__(self, evaluation_metric=None, sample_length=50):
+  def __init__(self, evaluation_metric=None, sample_length=50, use_baseline=True):
     self.sample_length = sample_length
     if evaluation_metric is None:
       self.evaluation_metric = xnmt.evaluator.BLEUEvaluator(ngram=4, smooth=1)
     else:
       self.evaluation_metric = evaluation_metric
-    self.regressor = None
+    self.baseline = None
+    self.use_baseline = use_baseline
 
   def __call__(self, translator, dec_state, src, trg):
     # TODO: apply trg.mask ?
@@ -284,16 +269,16 @@ class TranslatorReinforceLoss(Serializable):
       done = list(six.moves.map(lambda x: x == Vocab.ES, sample))
       if all(done):
         break
-    h_t = dy.tanh(translator.decoder.context_projector(dy.concatenate([dec_state.rnn_state.output(), dec_state.context])))
-    if not self.regressor:
-      self.regressor = self.Regressor(h_t.dim()[0][0])
 
-    decode_hidden_state = h_t.value()
-    dim = h_t.dim()
-    pred = self.regressor.predict_score(decode_hidden_state, dim)
+    if self.use_baseline:
+      h_t = dy.tanh(translator.decoder.context_projector(dy.concatenate([dec_state.rnn_state.output(), dec_state.context])))
+      if not self.baseline:
+        model = translator.decoder.param_col
+        self.baseline = linear.Linear(input_dim=h_t.dim()[0][0], output_dim=1, model=model)
+      self.bs = self.baseline(dy.nobackprop(h_t))
 
     samples = np.stack(samples, axis=1).tolist()
-    eval_score = []
+    self.eval_score = []
     for trg_i, sample_i in zip(trg, samples):
       # Removing EOS
       try:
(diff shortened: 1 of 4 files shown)


xnmt/translator.py
```diff
@@ -236,12 +237,33 @@ def __call__(self, translator, dec_state, src, trg):
...
+      self.param_collection = dy.Model()
+      self.trainer = dy.AdamTrainer(self.param_collection)
+      self.pW = self.param_collection.add_parameters((1, dim))
+      self.pb = self.param_collection.add_parameters(1)
+
+    def predict_score(self, h_t, dim):
+      h_t = dy.inputTensor(np.reshape(h_t, (dim[0][0], dim[1])), batched=True)
+      W = dy.parameter(self.pW)
+      B = dy.parameter(self.pb)
+      output = B + W * h_t
```
philip30: Use linear.Linear instead

xnmt/translator.py
```diff
@@ -236,12 +237,33 @@ def __call__(self, translator, dec_state, src, trg):
 class TranslatorReinforceLoss(Serializable):
   yaml_tag = '!TranslatorReinforceLoss'
 
+  class Regressor():
+    def __init__(self, dim):
+      self.param_collection = dy.Model()
+      self.trainer = dy.AdamTrainer(self.param_collection)
```
philip30: Not sure if adding 1 more trainer here is the correct way to incorporate the baseline. See https://github.com/neulab/xnmt/blob/segmenting_encoder/xnmt/segmenting_encoder.py#L211 for example. The idea is to compose the loss of the decoder and baseline together.

xnmt/translator.py
```diff
@@ -236,12 +237,33 @@ def __call__(self, translator, dec_state, src, trg):
...
+      self.pW = self.param_collection.add_parameters((1, dim))
+      self.pb = self.param_collection.add_parameters(1)
+
+    def predict_score(self, h_t, dim):
+      h_t = dy.inputTensor(np.reshape(h_t, (dim[0][0], dim[1])), batched=True)
```
philip30: use dy.nobackprop() instead to create a copy & non backpropagate-able graph

xnmt/translator.py
```diff
@@ -236,12 +237,33 @@ def __call__(self, translator, dec_state, src, trg):
 class TranslatorReinforceLoss(Serializable):
   yaml_tag = '!TranslatorReinforceLoss'
 
+  class Regressor():
```
philip30: Regressor can be implemented by a single Linear Transformation object located in linear.Linear.

xnmt/translator.py
```diff
@@ -277,9 +307,16 @@ def __call__(self, translator, dec_state, src, trg):
       except ValueError:
         pass
       # Calculate the evaluation score
-      eval_score.append(self.evaluation_metric.evaluate_fast(trg_i.words, sample_i))
+      score = self.evaluation_metric.evaluate_fast(trg_i.words, sample_i)
+      eval_score.append(0 if math.isnan(score) else score)
```
philip30: Do not squash nan to 0, this may create a more severe bug in the future I think.

xnmt/translator.py
```diff
@@ -277,9 +307,16 @@ def __call__(self, translator, dec_state, src, trg):
...
+
+    true = dy.inputTensor(eval_score, batched=True)
+    loss_exp = self.regressor.calc_loss(decode_hidden_state, eval_score, dim)
+    loss_exp.backward()
+    self.regressor.trainer.update()
```
philip30: There shouldn't be any training update or backward in calc_loss function

psjanani: Oh, I thought this would train the Regressor according to the inputs. How do you suggest that t=should be done?

xnmt/translator.py
psjanani: Makes sense. Currently, this nan is thrown when the target sentence is of zero length. I can test that before.

xnmt/translator.py
psjanani: Okay, got it. I shall look into the segmenting Encoder as suggested by you. Thanks.

xnmt/translator.py
```diff
@@ -277,9 +292,23 @@ def __call__(self, translator, dec_state, src, trg):
...
+      self.eval_score.append(score)
+      self.trueScore = dy.inputTensor(self.eval_score, batched=True)
+
+    final_score = -self.trueScore
+    if self.use_baseline:
+      final_score = self.bs - self.trueScore
+    return dy.sum_elems(dy.cmult(final_score , dy.esum(logsofts)))
+
+  @recursive_sum
+  def calc_additional_loss(self, _):
```
philip30: I don't think you need to do this. This is required if you are calculating the reward. In this case, the loss can be composed inside the __call__ and the LossBuilder() object is returned instead.

xnmt/translator.py
```diff
@@ -262,8 +269,16 @@ def __call__(self, translator, dec_state, src, trg):
       done = list(six.moves.map(lambda x: x == Vocab.ES, sample))
       if all(done):
         break
+
+    if self.use_baseline:
+      h_t = dy.tanh(translator.decoder.context_projector(dy.concatenate([dec_state.rnn_state.output(), dec_state.context])))
+      if not self.baseline:
+        model = translator.decoder.param_col
```
philip30: Initiate this in the constructor

xnmt/translator.py
```diff
@@ -277,9 +292,23 @@ def __call__(self, translator, dec_state, src, trg):
...
       # Calculate the evaluation score
-      eval_score.append(self.evaluation_metric.evaluate_fast(trg_i.words, sample_i))
-
-    return -dy.sum_elems(dy.cmult(dy.inputTensor(eval_score, batched=True), dy.esum(logsofts)))
+      score = 0 if not len(sample_i) else self.evaluation_metric.evaluate_fast(trg_i.words, sample_i)
+      self.eval_score.append(score)
+      self.trueScore = dy.inputTensor(self.eval_score, batched=True)
+
+    final_score = -self.trueScore
+    if self.use_baseline:
```
philip30: I think this is a bit different than I know. The baseline should be predicted at every timestep and be multiplied with the log prob at that time step, right? @neubig

xnmt/translator.py
philip30: Find a better way to pass the information about the size of the input. But this should be done in the constructor I think.

xnmt/translator.py
```diff
@@ -74,7 +78,8 @@ def __init__(self, src_embedder, encoder, attender, trg_embedder, decoder, loss_
       self.loss_calculator = TranslatorMLELoss()
     else:
       self.loss_calculator = loss_calculator
-
+    if (issubclass(self.loss_calculator.__class__, HierarchicalModel)):
+      self.register_hier_child(self.loss_calculator)
```
philip30: No need to check this. We can assume that the subclass is a HierarchicalModel

xnmt/translator.py
psjanani: Oh sorry, I initially read the algorithm and assumed that the word level predictions would be time-consuming. (I thought that the loss for Linear Regression would be according to BLEU score at every t-1th step which seemed too costly and ambiguous). 
So the intention was to develop the pipeline for sentence level approximation and then get to the word level in the next commit. But it seems to require just overall BLEU Score which makes complete sense and is not slow. Shall change it in next iteration. Thanks.

xnmt/translator.py
psjanani: Oh, I am confused here, @philip30. Are you suggesting to make the standard-loss an instance of LossBuilder? or am I misunderstanding something? Currently, I guess xnmt_train the "standard loss" (from the model's calc_loss) is considered by default to be a single loss and all additional losses seem to go as an instance of LossBuilder.  Wouldn't further generalizing this format make things more confusing?

xnmt/translator.py
philip30: Ah yes.. You are correct. I think this is true if we are using the reward for the whole sequence :) Please pardon my previous comment.

xnmt/translator.py
philip30: You can do this to substitute the return in __call__:
``` python
loss = LossBuilder()
loss.add_loss("reinf_loss", dy.sum_elems(dy.cmult(final_score , dy.esum(logsofts))))
loss.add_loss("baseline", dy.sum_batches(dy.squared_distance(self.trueScore, self.bs))
return loss
```
It doesn't matter if you return a loss_builder or an instance of dynet_expression as far as I know.

psjanani: Hi thanks a lot for you quick replies.
When I do this, I get the following error:
```     
   File "/xnmt/xnmt/xnmt_train.py", line 231, in run_epoch
   additional_loss = self.model.calc_additional_loss(dy.nobackprop(-standard_loss))   
   TypeError   :    bad operand type for unary -: 'LossBuilder'  
```

So, it seems that main algorithm doesn't expect LossBuilder object there. I can change it. But I wanted to ask if it is a good design choice.

philip30: I see. This is a new case, but I think the calc_loss should handle both the LossBuilder and dynet expression object. You can fix it with a design you think is good.
I think you can either implement the unary operator in LossBuilder or add a conditional statment there.

philip30: Or if it is too difficult, you can neglect the LossBuilder. Instead just use the "+" to sum 2 dynet expression. LossBuilder is used only for reporting purposes (so we can see 2 losses are going down separately)

xnmt/translator.py
psjanani: Hi!  I guess your initial observation seems to be valid. I might be wrong, but by adding the projected score (the projection of the final BLEU score) at every instance, we might be able to penalize the sentences that deviated early on, more.

PR #209, opened 2017-10-14
fix decoder input feeding dimension

cindyxinyiwang:
For input feeding, we should add the dimension of the context, not the dimension of the lstm


philip30:
I think this can be merged directly

neubig:
Yes, looks good! FYI: @philip30 and @msperber , I think you're familiar enough with the code now, so feel free to do review and merge PRs from other people.

PR #212, opened 2017-10-17
batcher breaks ties randomly

msperber:


philip30:
I think it is just a precision error @neubig . The numbers are really close

neubig:
Yes. But the test breaks, so we need to fix it.

I'm not sure whether the better solution is to ensure that batching is the same across the training and dev sets for the purpose of this test, or to ignore small differences between the numbers. I think it might actually be better to do the former, just to make sure that we get exactly the same number and aren't missing small differences. I could be convinced that the latter is OK though (and it's easier).

msperber:
Yes I also think it’s better to disable randomness for the test. I’ve encountered problems before that only made a very small difference. I’ll take a look tomorrow.

PR #213, opened 2017-10-18
fix train-dev-loss test; yaml-ize batcher

msperber:
I fixed the test by turning of randomization. In the process I also yaml-ized the Batchers. One caveat is that this changes the default to not use any batching, whereas previously the default was SrcBatcher with batch size 32. I should probably undo that change, but noticed that this will make it impossible to reach the "non minibatch" training mode, and I'm wondering why XNMT has that at all and whether it will be safe to remove it? Can't we just set minibatch size to 1 instead? @neubig


msperber:
This should be ready to merge, PTAL

PR #219, opened 2017-10-26
Debugging Reinforce

psjanani:
This change stabilizes Reinforce without Baseline.
Reinforce with Baseline is yet to be stabilized.


neubig:
LGTM!

Issue #153, opened 2017-08-03
Better Support for Sharing Components, Multi-task Learning


neubig:
Currently it is very difficult to implement multi-task learning in xnmt. This could probably be fixed by doing a few things:
1) Making it possible to define something like CompoundModel, which can contain a Translator and a Retriever, two Translators, etc.
2) Making it easier to "reference" previously defined model components. For example, Translator number two might "reference" the Encoder of Translator number 1. This would allow them to share a single encoder and train them in a multi-task fashion.
3) Come up with a new `TrainingTask` interface that references a model and its training data and parameters, a `DecodingTask` that performs decoding, and a `EvaluationTask` that performs evaluation for the various tasks.

This would be a large refactoring of the code, but could potentially make things much more flexible, so it would potentially be nice to have.

neubig:
(FYI: @msperber is working on this)

PR #222, opened 2017-10-27
Refactoring: Transducer

msperber:
- Remove the encoder concept for now, and replace it by transducers / sequence transducers (see transducer.py). These should be more flexible and modular. encoder.py is removed and the corresponding classes moved to lstm.py, pyramidal.py, conv.py, residual.py, ff.py. We also don't require 2 classes per encoder anymore.
- Remove the hierarchical model concept, which was mostly used to propagate events, and replace it with a global event handling mechanism (see events.py). This frees components from having to keep track of their sub-components, instead components can handle events themselves and e.g. enable/disable dropout without their parent component knowing about it.

This is a first step toward more flexibility needed for multi task training. Because changes are fairly invasive, I thought I should divide this into smaller, more manageable PRs. Please let me know what you think!


msperber:
Seems test_segmenting is failing because some resource is not set, probably due to switching from hierarchical model to the events system. It's hard for me to understand how exactly the events would need to be propagated in the segmenting encoder, but I'm guessing it would be a quick fix for @philip30 , maybe you can take a look?

philip30:
I actually would like the old encoders to be preserved. This fix needs me to change like 30 yaml files of my experiments.. And I think we need to still support the backward compatibility.

neubig:
@philip30 I understand, but personally I think xnmt has not evolved to the point where we should worry about preserving backward compatibility yet, especially if it ties our hands to sub-optimal design. And you can write a machine translation system to translate your old yaml files into new ones :)

One of the motivations behind this change is it will make it easier to implement your PR here in a clean way: https://github.com/neulab/xnmt/pull/191

philip30:
@msperber I think you need to pass the results of the parent event handling to the child. As far as I see the codes in the event.py, it is not handled yet. Actually they are building the html tree together (parent + childrent), and that was the original motivation of the hierarchical model. Maybe we can add the hierarchical event handling?

philip30:
It is not an easy fix btw

msperber:
@philip30 do we need control over the order in which event handlers are called? That's the main thing that changed I think. I did implement @register_xnmt_event_assign pretty much the same way you had implemented @recursive_assign, the only difference is that we can no longer be sure that the parent handler is called before the child handler, instead, the order is determined by the order in which objects are created. I think it should be possible to work with that, although I don't completely understand how the feature that got broken works.

The motivation for removing hierarchical model is that it will break when using component sharing, which is essential for supporting multitask learning as in #153. In that case, we no longer have a strict tree hierarchy. I understand it requires changing some existing code, but it will make things much more flexible. After this, changes regarding #153 would mostly be in the training procedure and serialization code, so I think the transducer code can mostly stay as it is after this PR.

msperber:
I was able to fix it, PTAL if this is ready to merge.

philip30:
@msperber thanks for fixing this. I took a look at the code briefly. I think it is OK to merge. 

Small comments: Please use 2 spaces tab in events.py

neubig:
Thanks for working this out! @philip30 and @msperber ! I also took a look and it looks OK fixing the spacing of tabs.

Issue #232, opened 2017-11-06
WordTrgSrcBatcher is not working in the current master branch.


philip30:
After merging master branch to my current working branch I found that there is a bug in the current implementation of WordTrgSrcBatcher. It is just broken and does not work. Has it been tested yet btw?

msperber:
I haven't used it and we only have unit tests for SrcBatcher and WordSrcBatcher so far, so I think it would be useful to add tests for the other batchers as well.

PR #233, opened 2017-11-07
Fix batcher

philip30:
Fix bugs in batcher.


philip30:
#232

philip30:
@neubig I don't understand why the test is failing. Can you help me to take a look?

msperber:
Hey Philip, it's possible this is an import error. Could you check if the attender is imported somewhere before the YAML parser is called? I had a similar issue before and don't understand it completely, but adding an import helped in my case.

philip30:
@msperber It runs OK in my environment, but it is just failing in travis-ci/pr. Did you have the same issue before? Thank you for the suggestion.

msperber:
Hmm no, I had the same error message, but I did have it in my local environment as well.

philip30:
I am merging it right now because it should run OK in most of the environments. This pr is only a bug fix also.

PR #236, opened 2017-11-10
Refactor training code

### LINKED ISSUE #223: masking not implemented correctly with initialization in lstm.py
in lstm.py, the customLSTMbuilder uses the class LSTMState. When you call add_input, you just pass in previous state. However, if you call a customLSTMbuilder with initial_state, it returns a LSTM state with some initialized c and h value. But this initial state does not have a previous_state property. So when you call add_input to it, it would not pass in the initialized c and h, and this initialization information would be lost forever?

msperber:
[X] Refactor
[X] resolve merge conflicts
[ ] Fix broken unit tests

This is a structural change to move toward #153 and also fixes #223.
- xnmt_train.py is no longer usable stand-alone (support was very limited previously anyways), is renamed to train.py; XnmtTrainer is now Serializable. In the future, we might use several XnmtTrainer instances for multi-task training.
- remove most usages of options, instead handling everything in a YAML-ized way. This does not affect the way config files are specified, but documentation is now implicit in the code; this was necessary for making XnmtTrainer Serializable
- trainers (Adam etc) are now Yaml objects and need to be configured accordingly.
- options parser is simplified: YAML is loaded as-is, no argument checking is performed, and the only special treatment of the "defaults" experiment is not to run it, but the actual copying over of settings needs to be done manually. This is very easy using standard YAML syntax, whereas the previous manual way of doing it seemed to complicated to maintain with multiple tasks. There is one new syntax element, kwargs, which is just a dictionary of init arguments used as __init__(**kwargs), which is useful in combination with the YAML << merging operator (e.g. see encoders.yaml config file).
- model-saving now saves the whole XnmtTrainer object. This allows removing the serialization_container, and loading / initialization from scratch share the same code (except that loading needs to populate dynet params in addition).
- it's now possible to overwrite parts of a pretrained model, although support is still a bit limited
- removed unused / buggy LSTM code

Here's a small example that illustrates how config file syntax changed:
```
defaults: &defaults # 'defaults' experiment is not run
  experiment:
    model_file: examples/output/<EXP>.mod
    # ...
  train: !XnmtTrainer
    batcher: !WordSrcBatcher
      avg_batch_size: 10
    glob: # global settings, passed through the whole model
      dropout: 0.1
      default_layer_dim: 512
    restart_trainer: True
    trainer: !AdamTrainer # now yaml-ized
      alpha: 0.0002
    corpus_parser: !BilingualCorpusParser
      src_reader: !PlainTextReader {}
      trg_reader: !PlainTextReader {}
      training_corpus: !BilingualTrainingCorpus # this moved inside the corpus parser
        train_src: examples/data/train.ja
        train_trg: examples/data/train.en
        dev_src: examples/data/dev.ja
        dev_trg: examples/data/dev.en
    model: !DefaultTranslator
      # no changes here
  decode:
    # ..
  evaluate:
    # ..

exp1:
  << : *defaults # copy over all settings from default experiment

exp2-finetune-model:
  << : *defaults
  train: !XnmtTrainer # load the pretrained model
    pretrained_model_file: examples/output/exp1-pretrain-model.mod
```


examples/component_sharing.yaml
```diff
@@ -8,27 +8,28 @@
...
 #                   this means the component is only created once, which also leads to DyNet parameters being shared.
 #                   Since YAML doesn't distinguish between a copy and a reference, we implement reference-sharing
-#                   by adding a unique __xnmt_id field to the component that is to be shared. During deserialization,
-#                   whenever a component's __xnmt_id is equal to the corresponding field of a component that has
+#                   by adding a unique _xnmt_id field to the component that is to be shared. During deserialization,
+#                   whenever a component's _xnmt_id is equal to the corresponding field of a component that has
 #                   has already been initialized, we re-use that initialized component instead of creating a new one.
 
-defaults:
+exp1.pretrain: &defaults
```
neubig: Should this be `&defaults` or `&exp1.pretrain`?
It might be worth thinking about what the canonical syntax here should be, as others will probably use it in the future, and I think perhaps giving the experiment the same name might be better?

examples/preproc.yaml
```diff
@@ -116,7 +116,3 @@ defaults:
     src_file: examples/output/test.norm.ja
   evaluate:
     ref_file: examples/output/test.norm.en
-
```
neubig: I like how we can simplify the syntax when we're running only one experiment.

examples/report.yaml
```diff
@@ -1,31 +1,32 @@
...
     err_file: examples/output/<EXP>.err
     run_for_epochs: 1
     eval_metrics: bleu
-  train:
-    trainer: adam
-    learning_rate: 0.01
-    default_layer_dim: 256 
-    dropout: 0.0
+  train: !XnmtTrainer
+    trainer: !AdamTrainer
```
neubig: The terminology `Trainer` is now overloaded. This is a bit annoying because, `AdamTrainer` should probably be `AdamOptimizer`, but that clashes with DyNet's terminology (which I don't plan on changing just because it'd be a mess trying to maintain backward compatibility). Maybe we can rename `XnmtTrainer` something like `TrainingRegimen`?

examples/component_sharing.yaml
msperber: Sure, that's a good idea, I'll try that!

examples/report.yaml
msperber: Sure, I can rename it to TrainingRegimen. xnmt.optimizer.AdamTrainer is currently a wrapper class around dynet.AdamTrainer (mostly because it seems not possible to serialize cython objects), so we could easily rename it to xnmt.optimizer.AdamOptimizer. What do you think? We also have training_strategy which is a bit confusing and could maybe be renamed..

msperber:
I'll merge for now but let me know if you'd like different names for xnmt.optimizer.AdamTrainer etc. and maybe training_strategy @neubig .

PR #238, opened 2017-11-15
Fixed tiny words/sec reporting bug

CharlotteKay:
The first words/sec reported in each epoch appears to be a bit smaller than others. The timer is not set correctly.


neubig:
Thanks!

PR #239, opened 2017-11-17
Added Label Smoothing

DevSinghSachan:
Hi Graham,

I have created this new branch in order to add the label smoothing code as there were a few merge conflicts in the older branch.

### DIFF
--- a/xnmt/decoder.py
+++ b/xnmt/decoder.py
@@ -44,7 +44,7 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
   def __init__(self, yaml_context, vocab_size, layers=1, input_dim=None, lstm_dim=None,
                mlp_hidden_dim=None, trg_embed_dim=None, dropout=None,
                rnn_spec="lstm", residual_to_output=False, input_feeding=True,
-               bridge=None):
+               bridge=None, label_smoothing=0.0):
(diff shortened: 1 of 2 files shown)


xnmt/decoder.py
```diff
@@ -119,13 +120,32 @@ def get_scores(self, mlp_dec_state):
...
+
     else:
-      return dy.pickneglogsoftmax_batch(scores, ref_action)
+      log_prob = dy.log_softmax(scores)
+      if not xnmt.batcher.is_batched(ref_action):
+        pre_loss = -dy.pick(log_prob, ref_action)
+      else:
+        pre_loss = -dy.pick_batch(log_prob, ref_action)
+
+      pre_loss = dy.mean_batches(pre_loss)
```
neubig: Why are you doing `mean_batches` here? I think you shouldn't do anything with respect to batches.

DevSinghSachan: actually, I got confused here and thought that the loss needs to be a scalar 1D value, so was doing mean_batches. I now realize that the loss needs to be present for every row.

I have updated the code now.

DevSinghSachan:
Hi Graham,
Removed the mean_batches() part, got confused earlier as to whether the loss needs to be 1D number or not.

Issue #240, opened 2017-11-21
PolynomialNormalization missing attribute


neilzxu:
c6fac72d109de8b7dffa2c6b15514ff4589565ef seems to have introduced a bug with PolynomialNormalization; can be reproduced by running the standard.yaml config file in examples/. 

Call:
`python3 xnmt/xnmt_run_experiments.py examples/standard.yaml --dynet-gpu`

Produces this output (truncated):

```
> Training
   Epoch 0.1002: train_loss/word=7.026294 (words=9426, words/sec=1150.65, time=0-00:00:08)
   Epoch 0.2000: train_loss/word=6.703589 (words=18864, words/sec=1155.63, time=0-00:00:16)
   Epoch 0.3003: train_loss/word=6.518545 (words=28110, words/sec=1188.03, time=0-00:00:24)
   Epoch 0.4005: train_loss/word=6.392840 (words=37588, words/sec=1138.11, time=0-00:00:32)
   Epoch 0.5003: train_loss/word=6.307682 (words=46648, words/sec=1159.72, time=0-00:00:40)
   Epoch 0.6004: train_loss/word=6.225853 (words=55790, words/sec=1144.99, time=0-00:00:48)
   Epoch 0.7000: train_loss/word=6.154869 (words=65143, words/sec=1154.20, time=0-00:00:56)
   Epoch 0.8003: train_loss/word=6.088884 (words=74466, words/sec=1143.59, time=0-00:01:04)
   Epoch 0.9000: train_loss/word=6.024372 (words=83640, words/sec=1188.11, time=0-00:01:12)
   Epoch 1.0000: train_loss/word=5.964533 (words=93086, words/sec=1154.51, time=0-00:01:20)
   Traceback (most recent call last):
     File "xnmt/xnmt_run_experiments.py", line 154, in <module>
          sys.exit(main())
     File "xnmt/xnmt_run_experiments.py", line 118, in main
          training_regimen.run_epochs(exp_args["run_for_epochs"])
     File "/home/ziyux/installs/miniconda3/envs/dynet/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/train.py", line 190, in run_epochs
          self.one_epoch()
     File "/home/ziyux/installs/miniconda3/envs/dynet/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/train.py", line 241, in one_epoch
          self.dev_evaluation()
[...]

PR #241, opened 2017-11-23
made xnmt_decode serializable

msperber:
This addresses #240. It's mostly a quick hack and not very pretty, but we can further refactor things later.


neubig:
Thanks a lot! It looked good enough to me, and having things not break is probably most important.

PR #246, opened 2017-11-29
Updated Transformer Code

DevSinghSachan:
This is the code from the paper "Attention is all you need".
I have tested this code on ja->en translation task in which it gets a BLEU score of around 25.5 on the dev set after 25 epochs.

Currently only greedy decoding is supported.


xnmt/translator.py
```diff
@@ -191,3 +194,183 @@ def html_report(self, context=None):
...
+    """
+    We convert the embeddings of masked input sequence to zero vector
+    """
+    (embed_dim, _), _ = embeddings.dim()
+    temp_mask = np.repeat(1. - mask[:, None, :], embed_dim, axis=1)
+    temp_mask = dy.inputTensor(np.moveaxis(temp_mask, [1, 0, 2], [0, 2, 1]), batched=True)
+    embeddings = dy.cmult(embeddings, temp_mask)
+    return embeddings
+
+  def initialize_position_encoding(self, length, n_units):
```
msperber: This looks like reusable code and could optionally be pulled out into a PositionalEmbedder that takes a base_embedder and adds the positional encoding on top.

xnmt/transformer.py
```diff
@@ -0,0 +1,321 @@
...
+    if sent_len == 1:
+        batch_A = dy.softmax(batch_A)
+    else:
+        batch_A = dy.softmax(batch_A, d=1)
+
+    # TODO: Check whether this is correct after masking
+    batch_A = dy.cmult(batch_A, mask)
+    assert (batch_A.dim() == ((n_querys, n_keys), batch * h))
+
+    # TODO: Check if attention dropout needs to be applied here
```
msperber: I think it does need to be applied on batch_A

xnmt/transformer.py
```diff
@@ -0,0 +1,321 @@
...
+    self.dropout = dropout
+
+  def __call__(self, e, xx_mask):
+    self.self_attention.set_dropout(self.dropout)
+    sub = self.self_attention(e, e, xx_mask)
+    e = e + dy.dropout(sub, self.dropout)
+    e = self.ln_1(e)
+
+    sub = self.feed_forward(e)
+    e = e + dy.dropout(sub, self.dropout)
```
msperber: One more thing: it would be better to add an if statement and skip the dropout operation for self.dropout==0.0

xnmt/translator.py
msperber: Actually, I was just looking into this because I need this also, and making a PositionalEmbedder seems not as easy as I thought because of param sharing, so we should leave this as is for now.

xnmt/batcher.py
```diff
@@ -112,7 +112,8 @@ def pack(self, src, trg):
     order = list(range(len(src)))
     return self.pack_by_order(src, trg, order)
 
-class ShuffleBatcher(Batcher):
+class ShuffleBatcher(Batcher, Serializable):
+  yaml_tag = u"!ShuffleBatcher"
```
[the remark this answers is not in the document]
DevSinghSachan: Yeah, I actually didn't notice that "SentShuffleBatcher" and "WordShuffleBatcher" classes are defined. I will remove the serialization part in here.

xnmt/transformer.py
```diff
@@ -0,0 +1,321 @@
...
+    return output
+
+
+class LayerNorm(object):
+  def __init__(self, dy_model, d_hid):
+    self.p_g = dy_model.add_parameters(dim=d_hid)
+    self.p_b = dy_model.add_parameters(dim=d_hid)
+
+  def __call__(self, input_expr):
+    # g = dy.parameter(self.p_g)
```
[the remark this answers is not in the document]
DevSinghSachan: @msperber. I noticed in my own code ("https://github.com/DevSinghSachan/aiayn") that, LayerNorm in dynet was not giving consistent BLEU scores. I wanted to check with you as to how was your experience of LayerNorm in self-attention models.

xnmt/transformer.py
```diff
@@ -0,0 +1,321 @@
...
+    batch_A = dy.cmult(batch_A, mask)
+    assert (batch_A.dim() == ((n_querys, n_keys), batch * h))
+
+    # TODO: Check if attention dropout needs to be applied here
+    batch_C = dy.transpose(batch_A * dy.transpose(batch_V))
+
+    # batch_C = batch_V * dy.transpose(batch_A)  # TODO: Check the correctness of this step
+    assert (batch_C.dim() == ((n_units // h, n_querys), batch * h))
+
+    C = dy.concatenate(split_batch(batch_C, h), d=0)
```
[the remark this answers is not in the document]
DevSinghSachan: Actually, I am not able to figure out if this is possible with reshape operations are not. 
I was trying concatenate for instance, batch 100 with batch 0, along the column dimension and am not sure if this feasible using reshape operations.

xnmt/transformer.py
DevSinghSachan: This is a good idea. I will do it in my future commits.

xnmt/transformer.py
```diff
@@ -0,0 +1,315 @@
...
+
+    batch_Q = dy.concatenate_to_batch(split_rows(Q, h))
+    batch_K = dy.concatenate_to_batch(split_rows(K, h))
+    batch_V = dy.concatenate_to_batch(split_rows(V, h))
+
+    assert(batch_Q.dim() == (n_units // h, n_querys), batch * h)
+    assert(batch_K.dim() == (n_units // h, n_keys), batch * h)
+    assert(batch_V.dim() == (n_units // h, n_keys), batch * h)
+
+    mask = np.concatenate([mask] * h, axis=0)
```
neubig: You shouldn't have to do this concatenation, just use broadcasting within DyNet. Also, what happens when `mask` is `None`?

xnmt/transformer.py
```diff
@@ -0,0 +1,315 @@
...
+    mask = dy.inputTensor(np.moveaxis(mask, [1, 0, 2], [0, 2, 1]), batched=True)
+    batch_A = (dy.transpose(batch_Q) * batch_K) * self.scale_score
+    batch_A = dy.cmult(batch_A, mask) + (1 - mask)*MIN_VALUE
```
neubig: This is inefficient. I'd add `MIN_VALUE` to mask before calling `inputTensor`.

xnmt/transformer.py
```diff
@@ -0,0 +1,315 @@
...
+    batch_C = dy.transpose(batch_A * dy.transpose(batch_V))
+    assert (batch_C.dim() == ((n_units // h, n_querys), batch * h))
+
+    C = dy.concatenate(split_batch(batch_C, h), d=0)
+    assert (C.dim() == ((n_units, n_querys), batch))
+    C = self.finishing_linear_layer(C)
+    return C
+
+
+class FeedForwardLayer(object):
```
neubig: I'm not sure if it is necessary to add this, but if it is necessary, it should be in a separate file such as `xnmt.mlp`.

xnmt/transformer.py
neubig: Yes, please do.

xnmt/translator.py
```diff
@@ -191,3 +194,190 @@ def html_report(self, context=None):
     return html
 
 
+class TransformerTranslator(Translator, Serializable, Reportable):
```
neubig: At some point, I'd like to remove the need for having this be a separate translator, but I think it's OK for now.

xnmt/embedder.py
```diff
@@ -55,7 +57,8 @@ def __init__(self, yaml_context, vocab_size, emb_dim = None, weight_noise = None
     self.weight_noise = weight_noise or yaml_context.weight_noise
     self.word_dropout = word_dropout
     self.fix_norm = fix_norm
-    self.embeddings = yaml_context.dynet_param_collection.param_col.add_lookup_parameters((self.vocab_size, self.emb_dim))
+    self.embeddings = yaml_context.dynet_param_collection.param_col.add_lookup_parameters((self.vocab_size, self.emb_dim),
+                                                                                          init=linear_init(self.vocab_size))
```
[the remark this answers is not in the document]
DevSinghSachan: I did several runs with and without this LeCun initialisation and found that with this initialisation, model's convergence improved in transformer model.

xnmt/loss_tracker.py
```diff
@@ -78,15 +80,15 @@ def report_train_process(self):
...
+      print(LossTracker.REPORT_TEMPLATE.format('train', learning_rate) % (
         self.fractional_epoch, self.epoch_loss.sum() / self.epoch_words,
         self.epoch_words,
         (self.epoch_words - self.last_report_words) / (this_report_time - self.last_report_train_time),
         self.format_time(time.time() - self.start_time)))
 
       if len(self.epoch_loss) > 1:
         for loss_name, loss_values in self.epoch_loss:
-          print("- %s %5.6f" % (loss_name, loss_values / self.epoch_words))
+          print("- %s %5.6f" % (loss_name, loss_values))
```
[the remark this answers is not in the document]
DevSinghSachan: Actually, I changed it for debugging purposes to account for the mean loss. I will revert it to the original formulation now.

xnmt/train.py
```diff
@@ -236,7 +237,7 @@ def one_epoch(self, update_weights=True):
         self.trainer.update()
 
       # Devel reporting
-      self.logger.report_train_process()
+      self.logger.report_train_process(self.trainer.optimizer.learning_rate)
```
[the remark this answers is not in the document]
DevSinghSachan: This argument is not important as such. I included it for debugging purposes to monitor the learning rate of the optimizer at every step.

xnmt/embedder.py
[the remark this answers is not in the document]
neubig: Sure, that's good. But this is in `embedder.py` so it will influence all models in xnmt. I think perhaps we should have this as an option. Then we can change the default when we're sure it's good in all models.

xnmt/transformer.py
```diff
@@ -0,0 +1,315 @@
...
+    self.is_self_attention = self_attention
+    self.attn_dropout = attn_dropout
+
+  def set_dropout(self, dropout):
+    self.dropout = dropout
+
+  def __call__(self, x, z=None, mask=None):
+    h = self.h
+    if self.is_self_attention:
+      Q = self.W_Q(x)
```
neubig: I took another look at this, and I still think that this is incorrect. Look at the equation at the end of Section 3.2.2: https://arxiv.org/pdf/1706.03762.pdf

It's not doing a single matrix multiply over the entire input, but instead `h` individual matrix multiplies over each of the segments of the input.

To connect what you're doing with what they're doing in the paper: your matrix is a regular matrix with all the parameters being non-zero. Their matrix is a block diagonal matrix (https://en.wikipedia.org/wiki/Block_matrix#Block_diagonal_matrices), where all of the parameters that are off the diagonal are zero.

So basically, I think you should implement attention similar to the formulas in their paper just using for loops for now. We could consider a more efficient way to do this with tensor contractions later.

xnmt/transformer.py
DevSinghSachan: Actually, this is a different type of broadcasting operation specifically for multi-head attention. I think without this, it would be infeasible to implement.
In my implementation, the mask will never be None. I define masks for for every step for both encoder and decoder

xnmt/transformer.py
DevSinghSachan: Actually, I was doing MLP over sentences using linear layer, so I have renamed it now to FeedForwardLayerSent

xnmt/transformer.py
DevSinghSachan: mask is used in two places, in lines 139 and 147. I am actually not sure how to make this efficient. 
I have tried a couple of ways to do masking, but did not get noticeable increase in speed.

xnmt/linear.py
```diff
@@ -1,12 +1,26 @@
 import dynet as dy
+from xnmt.initializer import LeCunUniform as linear_init
+
 
 class Linear(object):
   def __init__(self, input_dim, output_dim, model):
-    self.W1 = model.add_parameters((output_dim, input_dim))
+    self.W1 = model.add_parameters((output_dim, input_dim), init=linear_init(input_dim))
```
neubig: This is again changing the default behavior of the entire xnmt toolkit without making it an option. Make sure these things are set as options.

xnmt/linear.py
```diff
@@ -1,12 +1,26 @@
...
     self.b1 = model.add_parameters(output_dim)
 
   def __call__(self, input_expr):
     W1 = dy.parameter(self.W1)
     b1 = dy.parameter(self.b1)
 
     return dy.affine_transform([b1, W1, input_expr])
+
+
+class LinearNoBias(object):
```
neubig: Actually, this doesn't need to be a separate class. Can't we just have a `bias` keyword argument for Linear?

xnmt/train.py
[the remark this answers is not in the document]
neubig: Why don't we pass in the trainer to logger at some point, so logger can save a reference to the trainer. Then if this reference is not None, logger will print out the learning rate in its logging information. I don't like adding an extra argument to this function, as it may make things complicated down the line.

xnmt/transformer.py
neubig: OK, let's keep it for now.

xnmt/translator.py
```diff
@@ -191,3 +194,190 @@ def html_report(self, context=None):
...
+    log_timescale_increment = (np.log(10000. / 1.) / (float(num_timescales) - 1))
+    inv_timescales = 1. * np.exp(np.arange(num_timescales).astype('f') * -log_timescale_increment)
+    scaled_time = np.expand_dims(position, 1) * np.expand_dims(inv_timescales, 0)
+    signal = np.concatenate([np.sin(scaled_time), np.cos(scaled_time)], axis=1)
+    signal = np.reshape(signal, [1, length, channels])
+    self.position_encoding_block = np.transpose(signal, (0, 2, 1))
+
+  def make_input_embedding(self, emb_block, length):
+    emb_block = emb_block * self.scale_emb
+    emb_block += dy.inputTensor(self.position_encoding_block[0, :, :length])
```
neubig: I think we should add a check here to see if length is greater than the length of `position_encoding_block`, and if so, re-call `initialize_positional_encoding(length*2, num_units)`

neubig: (I had a crash because I had a sentence larger than 500)

xnmt/batcher.py
```diff
@@ -112,7 +112,7 @@ def pack(self, src, trg):
     order = list(range(len(src)))
     return self.pack_by_order(src, trg, order)
 
-class ShuffleBatcher(Batcher):
+class ShuffleBatcher(Batcher, Serializable):
```
msperber: Could you remove the Serializable here? It might trick people into thinking they can specify the ShuffleBatcher in the config file.

xnmt/train.py
```diff
@@ -322,8 +326,10 @@ def dev_evaluation(self, out_ext=".dev_hyp", ref_ext=".dev_ref", encoding='utf-8
   def compute_dev_loss(self):
     loss_builder = LossBuilder()
     trg_words_cnt = 0
+    steps = 0
```
msperber: It seems steps are never used, remove? The same holds in the logger.

msperber:
The merge conflicts come from the logging of the learning rate. I'm wondering if we should remove those parts of the code, and instead have the TransformerAdamTrainer print the logs, e.g. after every n updates? That would be less intrusive in terms of code and make merging much easier.

neubig:
Thanks a lot for dealing with all the suggestions! Three final things, could you:

1. Add an integration test to the unit tests. This can be done by adding a config file to `test/config/`, then adding a test to `test/test_run.py`.
2. Fix Matthias's comment about dead code.
3. Fix the merge conflicts.
4. Add a comment that support for this is somewhat experimental. We haven't yet completely reproduced the paper's experimental results, so I'd like to mark it as such until we do.

Once these are done I think we're ready to merge!

DevSinghSachan:
Thanks for the comments Graham and Matt.
I have tried to remove the optimizer logging from the train.py and loss_tracker.py but, it is still giving merge conflicts.
I am not quite sure as to why merge conflicts are coming.

msperber:
It looks like just a couple of lines, you can run ```git merge master``` and then fix those lines and push.

DevSinghSachan:
Thanks a lot. Yes, it fixed the merge conflict. But, I had to delete train.py file.

neubig:
Looks like tests are failing. Could you try running tests with `python -m unittest discover`?

DevSinghSachan:
Hi Graham, Matthias,
I have fixed the test case now and made other changes as well according to the new merged code in xnmt master.
There were some merge issues which I couldn't resolve, So I created a new pull request here:
https://github.com/neulab/xnmt/pull/261

neubig:
Closing this because it has been merged in another branch.

Issue #220, opened 2017-10-26
Implement External Evaluator


neubig:
It would be nice to be able to call an external evaluation program. cdec has a very nice interface for doing so:
> https://github.com/redpony/cdec/tree/master/mteval

It would be cool if we had something like this.

neubig:
Done by #247

PR #247, opened 2017-12-02
Added support for an external evaluation script

roeeaharoni:
This pull request adds support for an external evaluation script, as supported in other NMT frameworks like Nematus and Marian. 

To use this feature, add: `external|path=PATH` to the `eval_metrics` configuration. The evaluation script in PATH should only print the evaluation score (and only it) before termination. 

In case a lower score is better, one should indicate it by setting `higher_better` to False, i.e.: `external|path=PATH|higher_better=False`

### DIFF
--- a/examples/eval.sh
+++ b/examples/eval.sh
@@ -1,0 +1,2 @@
+#!/usr/bin/env bash
+echo 30.00
--- a/examples/eval.sh
+++ b/examples/eval.sh
@@ -1,2 +1,0 @@
-#!/usr/bin/env bash
-echo 30.00
--- a/examples/standard.yaml
+++ b/examples/standard.yaml
@@ -1,6 +1,5 @@
 standard:
   experiment:
-    external_eval_script: examples/eval.sh
     model_file: examples/output/<EXP>.mod
     hyp_file: examples/output/<EXP>.hyp
     out_file: examples/output/<EXP>.out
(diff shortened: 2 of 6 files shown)


neubig:
Thanks! But this re-opens the pipe every time a sentence is evaluated, right? This seems slow. Perhaps this could be reconciled with this issue? The interface specified there would allow the pipe to be open once upon initialization of the evaluator, then used over and over again: https://github.com/neulab/xnmt/issues/220

roeeaharoni:
As far as I understood the current implementation, the evaluator's evaluate() is called once for the entire test set and not per sentence, so only one pipe will be opened in each evaluation. Note that the external script does not directly receive the outputs: it can either use the ones already written to a file by train.py before evaluation or decode them by itself.

neubig:
@roeeaharoni Yeah, I guess that's right for the moment. We'll probably have to refactor a bit to do what I suggested, so I'll merge this PR first then we can go back and fix things later. Thanks!

msperber:
This is currently breaking tests under Python 3.6, could you fix this @roeeaharoni ? It's probably just a matter of adding brackets to the print statements.

roeeaharoni:
@msperber Added the missing print brackets in my fork (roeeaharoni/xnmt, xnmt_evaluate.py, line 66). I don't see the new commit in the pull request (probably since it is already merged) - should I create another pull request with the fix?

msperber:
Sure, thanks!

roeeaharoni:
@msperber done, see: https://github.com/neulab/xnmt/pull/252

Issue #234, opened 2017-11-08
preproc.yaml should include SentencePiece detokenization


neubig:
I believe this is missing detokenization when evaluating BLEU score: https://github.com/neulab/xnmt/blob/master/examples/preproc.yaml

john-hewitt:
A grep showed that `detokenize` is never called anywhere in the pipeline, though preproc.py is set to be able to call the SentencePiece tokenizer. So, it's likely more than just a config file problem. Would such a call go somewhere like [here](https://github.com/neulab/xnmt/blob/c6fac72d109de8b7dffa2c6b15514ff4589565ef/xnmt/xnmt_decode.py#L118), in the decoder output processor? Or in `xnmt_evaluate.py`?

neubig:
Yes, I think that function is what you want, and it probably just needs to be added to the config file. Thanks!

neubig:
Fixed by #248

Issue #244, opened 2017-11-27
Checking string instance different between Python2/Python3 in tokenization


neilzxu:
Reproduce
`python3 ~/src/xnmt/xnmt/xnmt_run_experiments.py examples/preproc.yaml`
with the appropriate paths substituted in for tokenizers.

Output:
```
Traceback (most recent call last):
  File "/home/ziyux/src/xnmt/xnmt/xnmt_run_experiments.py", line 154, in <module>
    sys.exit(main())
  File "/home/ziyux/src/xnmt/xnmt/xnmt_run_experiments.py", line 84, in main
    xnmt.xnmt_preproc.xnmt_preproc(**preproc_args)
  File "/home/ziyux/src/xnmt/xnmt/xnmt_preproc.py", line 47, in xnmt_preproc
    in_stream = tokenizer.tokenize_stream(in_stream)
  File "/home/ziyux/src/xnmt/xnmt/preproc.py", line 95, in tokenize_stream
    tokenized_string = self._tokenize(stream.read())
  File "/home/ziyux/src/xnmt/xnmt/preproc.py", line 160, in _tokenize
    if isinstance(string, unicode):
NameError: name 'unicode' is not defined
```

neubig:
Should be fixed by #248

PR #248, opened 2017-12-03
fixing preprocessing tokenization

neubig:
OK, I started fixing https://github.com/neulab/xnmt/issues/234

Basically, I've figured out the problems:
- [x] In `preproc.yaml`, we need to add the `join-char` preprocessing step in `xnmt_decode`
- [x] We need to specify sentencepieced input and output in the yaml file
- [x] We need to fix the join-char step to treat unicode underscores as the thing to merge together (which is the sentencepiece default). This is done by default now, but for some reason is not working properly.

The following are things that we're probably not going to handle in this PR.

* Ideally, we'd like to be able to specify this separator in the config file, probably by passing in the output strategy as a class using the `!ClassName` syntax
* There is a problem with evaluation of likelihood and BLEU at dev time. Tokenized output files are necessary for evaluation of likelihood, but untokenized output files are necessary for evaluation of BLEU. Right now we can't specify both, so basically we need to use untokenized files and just ignore the poor likelihood scores.


neubig:
Note: this also started to address #244 as well.

john-hewitt:
I'd be happy to contribute post-Dec 15. I'll plan on it unless you're planning to get it done before then.

neubig:
Added `join-piece` via https://github.com/neulab/xnmt/pull/255
And we probably don't need to specify the separator for this PR.

neubig:
Most recent updates fixed https://github.com/neulab/xnmt/issues/249

john-hewitt:
I've been testing this on a new Python 3.6 install and the encoding choice code in preproc.py is still giving errors for me. Will update later tonight.

john-hewitt:
The directory to write the sentencepiece model to wasn't being created before sentencepiece tried to write the model. Error messages were then being suppressed by an unrelated condition. Fixed in #262.

PR #250, opened 2017-12-03
Gleu Score

psjanani:
Implemented GLEU score evaluation.

Test: The scores were compared with those of nltk implementation for ~5000 pairs of sentences at sentence and corpus level.
And its integration with xnmt was tested.


msperber:
LGTM, I'll merge this!

PR #251, opened 2017-12-04
Multi task training

msperber:
Multi-task training seems to be working. Needs a bit more testing and cleanup, but I wanted to ask for feedback about the interface. Main places to look at would be training_regimen.py, training_task.py, speech_multi.yaml

#153


test/config/speech_multi.yaml
```diff
@@ -0,0 +1,88 @@
...
+    err_file: examples/output/<EXP>.err
+    eval_metrics: cer,wer
+    save_num_checkpoints: 2
+  train: !JointMultiTrainingTask
+    tasks:
+    - !TrainingRegimen
+      name: first_task
+      run_for_epochs: 2
+      src_format: contvec
+      trainer: !AdamTrainer &task1_trainer
```
neubig: It's not clear to me why we need two IDs here?

xnmt/training_regimen.py
```diff
@@ -0,0 +1,325 @@
...
+            DependentInitParam(param_descr="model.src_embedder.vocab", value_fct=lambda: initialized_subcomponents["corpus_parser"].src_reader.vocab),
+            DependentInitParam(param_descr="model.trg_embedder.vocab", value_fct=lambda: initialized_subcomponents["corpus_parser"].trg_reader.vocab)]
+
+  def pack_batches(self):
+    self.train_src, self.train_trg = \
+      self.batcher.pack(self.corpus_parser.get_training_corpus().train_src_data, self.corpus_parser.get_training_corpus().train_trg_data)
+    self.dev_src, self.dev_trg = \
+      self.batcher.pack(self.corpus_parser.get_training_corpus().dev_src_data, self.corpus_parser.get_training_corpus().dev_trg_data)
+
+  def _augment_data_initial(self):
```
neubig: I wasn't able to understand this. Add a comment?

xnmt/training_task.py
```diff
@@ -0,0 +1,178 @@
...
+    loss.backward()
+    trainer.update()
+  
+  def checkpoint_needed(self):
+    raise NotImplementedError()
+
+  def checkpoint(self, out_ext=".dev_hyp", ref_ext=".dev_ref", encoding='utf-8'):
+    raise NotImplementedError()
+
+class BaseMultiTrainingTask(BaseTrainingRegimen, TrainingTask):
```
neubig: Maybe add a short comment?

xnmt/training_regimen.py
```diff
@@ -0,0 +1,325 @@
...
+import xnmt.xnmt_decode
+import xnmt.xnmt_evaluate
+from xnmt.evaluator import LossScore
+import xnmt.optimizer
+from xnmt.events import register_xnmt_event
+'''
+This will be the main class to perform training.
+'''
+
+class TrainingRegimen(xnmt.training_task.BaseTrainingRegimen, xnmt.training_task.TrainingTask, Serializable):
```
neubig: Comments on all the functions in this class would aid understanding.

msperber:
Thanks for taking a look, I added several comments.

One thing that I'm unsure about is whether division into ```BaseTrainingRegimen``` (implements a training loop) and ```TrainingTask``` (implements training steps) is confusing. Both single- and multi-task training regimens implement both, but the tasks of the multi-task regimens is only required to implement the latter.

msperber:
Other than the tests failing for some reason this would be ready unless there are concerns. Besides refactoring and implementing the multi-task features, I also renamed some components for better consistency and less terminology overlap. Here's an overview:
- ```TrainingRegimen``` is a base class that implements a training loop
- ```TrainingTask``` is a base class that implements training steps, minibatch generator, etc.
- ```SimpleTrainingRegimen``` (previously ```TrainingRegimen```) implements both, with methods modularized accordingly
- The multi task training regimens contain a list of tasks and implements its own training loop while invoking methods from the ```TrainingTask``` interface for each tasks
- Multi task training always assumes the first task being the main task that controls the optimizer, training schedule, model checkpoints, etc.
- ```XnmtDecoder```, ```xnmt_decoder```, ```decode```: renamed to ```SimpleInference```, ```simple_inference```, ```inference```: to avoid terminology overlap with the “decoder” part of an encoder-decoder model

Tests fail under 3.6 with a segmentation fault. It works on my machine with both Python 2.7 and 3.6. Any ideas what could be going wrong here?

test/test_training.py
```diff
@@ -38,7 +38,7 @@ def assert_single_loss_equals_batch_loss(self, model, batch_size=5):
     Tests whether single loss equals batch loss.
     Truncating src / trg sents to same length so no masking is necessary
     """
-    batch_size = 5
+    batch_size = batch_size
```
neubig: Remove?

xnmt/optimizer.py
```diff
@@ -48,3 +48,19 @@ class AdamTrainer(XnmtOptimizer, Serializable):
   def __init__(self, yaml_context, alpha = 0.001, beta_1 = 0.9, beta_2 = 0.999, eps = 1e-8):
     self.optimizer = dy.AdamTrainer(yaml_context.dynet_param_collection.param_col, 
                                     alpha, beta_1, beta_2, eps)
+
+class LazySGDTrainer(XnmtOptimizer, Serializable):
```
neubig: Is this limited to SimpleSGDTrainer only? It'd be nice if we could do some `args`+`kwargs` magic to make it support any trainer that we want. This could be a future commit though.

msperber: This is only needed for the default trainer which is currently SGD. The main task can use any trainer it wants, and the auxiliary tasks use the lazy trainer which is not instantiated because only the main task's trainer is used. Not sure if this answers your question/suggestion?

neubig: Actually, I'm confused. Why don't we use trainers for the auxiliary tasks? Their parameters need to be updated as well, right?

msperber: Actually, all tasks' parameters are in the same parameter collection, which is updated with the same trainer. So one trainer updates the parameters of all tasks. I'm not sure if it's possible to have individual trainers for different parameter sub-collections, but even if it is that would probably confusing when we start sharing parameters across tasks?

neubig: It is possible to have different trainers for different sub-collections (I think). I'm not super-happy with a design where we basically introduce a class that doesn't do anything as a band-aid to prevent our program from crashing. Do you think it'd be worth changing where the trainer lives in the code base so we don't have to do this patch? Perhaps make it a property of MultiTaskTrainingRegimen, in which case this trainer would be used by all sub-TrainingRegimens?

msperber: Yeah, I agree this is a bit hacky. The problem is that SimpleTrainingRegimen also have trainers that are instantiated to the default if set to None, and they know nothing about whether they live inside a multi-task training regimen or not. A few ways to solve this might be:
- pull trainer out of the "train" part and into the "experiments" part, although that seems a bit counter-intuitive
- make the SGDTrainer lazy. That way we have a bit of band-aid code in the SGDTrainer, but at least no extra class definition.
- (maybe my favorite) check during deserialization that only one trainer is instantiated and create a reference to this trainer if more trainers are specified. This can probably be done by using the _xnmt_id mechanism also used for parameter sharing.

What do you think?

neubig: Hmm, this is also not the absolute best solution, but how about if SimpleTrainingRegimen's trainer is None we don't instantiate it in the constructor, but instead wait until the trainer is used. This is the same thing that LazySGDTrainer is doing, but it seems to make more sense to me and doesn't require the extra class definition or adding too much complexity.

msperber: I arrived at a solution now by breaking apart SimpleTrainingRegimen into two classes:
- SimpleTrainingTask has no trainer and is used inside the multi-task regimens
- SimpleTrainingRegimen extends it by holding a trainer and specifying the training loop
This allows keeping only one trainer that lives in the top of the hierarchy, as you suggested first.

msperber:
I'm running two test training runs right now to make sure I didn't break anything. I'll merge once those are done.

msperber:
Everything looks fine, so I'll merge this!

PR #252, opened 2017-12-04
Added support for an external evaluation script

roeeaharoni:
Identical to this pull request: https://github.com/neulab/xnmt/pull/247
But now with python3-style prints.


neubig:
This is re-doing all the previous commits. Could you please create a new PR that only fixes the bracketing?

roeeaharoni:
Done, see here: https://github.com/neulab/xnmt/pull/253
Sorry for the mess!

neubig:
I think this should be fixed by #253. Please re-open a new PR if you have something else that'd be useful to add.

PR #253, opened 2017-12-04
added bracketing to print for python3 support

roeeaharoni:
added bracketing to print for python3 support in external evaluation script implementation


neubig:
Thanks!

PR #254, opened 2017-12-05
Embed sent

neubig:
I think having the for-loop like behavior as a default for sentence embedders seems reasonable.


neubig:
Woops, let me revert the preprocessing part.

neubig:
OK, should be ready for review now.

PR #255, opened 2017-12-05
desentencepice functionality

cindyxinyiwang:


xnmt/output.py
```diff
@@ -60,3 +60,14 @@ def __init__(self, merge_indicator=u"@@"):
 
   def words_to_string(self, word_list):
     return u" ".join(word_list).replace(self.merge_indicator_with_space, u"")
+
+class JoinedPieceTextOutputProcessor(PlainTextOutputProcessor):
+  '''
```
neubig: This comment is wrong?

xnmt/output.py
```diff
@@ -60,3 +60,14 @@ def __init__(self, merge_indicator=u"@@"):
 
   def words_to_string(self, word_list):
     return u" ".join(word_list).replace(self.merge_indicator_with_space, u"")
+
+class JoinedPieceTextOutputProcessor(PlainTextOutputProcessor):
```
neubig: I don't think we need a separate JoinedPieceTextOutputProcessor, as it's basically the same as JoinCharTextOutputProcessor.

xnmt/xnmt_decode.py
```diff
@@ -146,6 +146,8 @@ def get_output_processor(self):
       return JoinedCharTextOutputProcessor()
     elif spec == "join-bpe":
       return JoinedBPETextOutputProcessor()
+    elif spec == "join-piece":
+      return JoinedPieceTextOutputProcessor()
```
neubig: Maybe we can add `merge_indicator=u"\u2581"`

xnmt/output.py
cindyxinyiwang: JoinCharTextOutput is slightly different, because it only replaces a single space token with space. For sentence piece, a word could have a space token attached to it at the beginning, so we need to do a replacement after merging

xnmt/output.py
cindyxinyiwang: I will update the comment

xnmt/output.py
neubig: Ahh, right. Thanks!

PR #257, opened 2017-12-19
Pad to multiple

msperber:
Some encoders that do downsampling require inputs to be multiple of the downsampling factor. This PR moves responsibility for padding-to-multiple to the batcher which I think is the most appropriate place for it because it also manages the padding token. Previously, every component that required padding-to-multiple had to implement its own solution.

At inference time, we're not using minibatches, but will still have to invoke the batcher to invoke padding-to-multiple. In the future it would be good to expand on this and use minibatches consistently even during inference (and simply set the batch size to 1).


msperber:
Done, I'll merge.

PR #258, opened 2017-12-21
Initial patience

msperber:
This implements ```initial_patience``` that allows delaying the first LR decay further than subsequent LR decays which are controlled by ```patience``` (previously ```attempts_before_lr_decay```).

This is useful when using a ```schedule_metric``` other than ```loss```, which often suffers from the optimization metric's high variance during the early training stages.


neubig:
Thanks!

PR #262, opened 2017-12-22
make sentpiece model dir; fix stderr preproc output

john-hewitt:
- sentencepiece models weren't being built for me on a clean install because the `preproc` folder wasn't being made before sentencepiece was initialized. 
 - This error was being hidden because stderr messages were only being written given an unrelated condition.
This PR fixes those bugs for Python 3.6. Python 2 should work but no promises.


neubig:
Thanks!

PR #266, opened 2017-12-27
add general linear attender of https://arxiv.org/abs/1508.04025

john-hewitt:
The _general_ attender of https://arxiv.org/abs/1508.04025 is a nice compromise between the more complicated MlpAttender and the unparametrized DotAttender. It is implemented here.

## quick benchmarks on examples/data/{train,dev,test}.{ja,en}
Summary of below -- GeneralLinear seems to be at least on par, BLEU-wise, with the MLP. May be better, but I have my doubts. The improvement may be due to its having fewer parameters, since the train set is so small. It's also considerably slower, which is unintuitive to me, but maybe someone can rewrite line 131 

### MlpAttender
- train words/sec=2574.60ish
- Dev bleu 0.016222765483504837 (20 epochs)
- Test bleu 0.02349701978746616 (20 epochs)

### GeneralLinearAttender
 - train words/sec=1540.73ish
 - Dev bleu 0.031945032510106094 (20 epochs)
 - Test bleu 0.04681154243162295 (20 epochs)

### DIFF
--- a/xnmt/attender.py
+++ b/xnmt/attender.py
@@ -105,12 +105,12 @@ class DotAttender(Attender, Serializable):
     I = self.curr_sent.as_tensor()
     return I * attention
 
-class GeneralLinearAttender(Attender, Serializable):
+class BilinearAttender(Attender, Serializable):
   '''
   Implements the general linear attention of https://arxiv.org/abs/1508.04025
   '''
 
-  yaml_tag = u'!GeneralLinearAttender'
+  yaml_tag = u'!BilinearAttender'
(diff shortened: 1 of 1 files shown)


neubig:
Thanks! I prefer the term "bilinear" though, as it's more common. Why don't we call it BilinearAttender, but keep a common saying that it's equivalent to the generalized linear attender in Luong et al.

john-hewitt:
Okay, agreed. Changed the name and made a comment in the docstring.

Issue #260, opened 2017-12-21
Disconnect between evaluation and processing


neubig:
There is a problem with evaluation of likelihood and BLEU at dev time when using output tokenized by, for example, sentence piece. Tokenized output files are necessary for evaluation of likelihood, but untokenized output files are necessary for evaluation of BLEU. Right now we can't specify both.

I would like to propose an overhaul about how we think about evaluation. Each type of evaluation (e.g. evaluation of perplexity, evaluation of MT metrics) each gets its own class. These evaluation tasks can take in different reference files. We don't include the detokenizer in the inference engine, and instead it lives in the evaluation task. These evaluation tasks could be shared between dev time and test time by using the component sharing mechanism if necessary.

What do you think @msperber ?

msperber:
Yes, it seems the evaluation approach could use some restructuring. I think that (after correctness obviously) one goal should be to keep the config files as simple as possible. I attached an example config, is that what you were thinking of? There are a couple of problems with this particular version though:
- it's difficult to avoid copying the evaluators between dev and test because they need different reference files
- some evaluators might need several post-processing steps (such as join-char and detokenize), how should we go about supporting that? Maybe reusing some of the infrastructure from preproc.py?
- to support the schedule_metric, instead of one 'evaluators' attribute we might to need to: 'schedule_evaluator' and 'auxiliary_evaluators'

```exp:
  experiment: ...
  train: !SimpleTrainingRegimen
    evaluators:
    - !BleuEvaluator
      post_process: join-char
      ref: dev.trg.words
    corpus_parser: !BilingualCorpusParser
      src_reader: !PlainTextReader {}
      trg_reader: !PlainTextReader {}
      training_corpus: !BilingualTrainingCorpus
        train_src: train.src
        train_trg: train.trg.char
        dev_src: dev.src
        dev_trg: dev.trg.char
    model: ...
  inference: !SimpleInference
    src_file: examples/output/test.norm.ja
  test-evaluators:
  - !BleuEvaluator
    post_process: join-char
    ref: test.trg.words
```

neubig:
OK, I've thought about this a bit, and here is a proposed model format, which I'll explain a bit below the fold.

    experiment: ...
      model:
        ... # model parameters
        inference: !SimpleInference
          beam: 5
        src_reader: !PlainTextReader {}
        trg_reader: !PlainTextReader {}
      train: !SimpleTrainingRegimen
        ... # training parameters
        src_file: train.src
[...]

PR #267, opened 2018-01-03
[WIP] Refactoring of evaluation

neubig:
This branch will attempt to refactor how we handle evaluation by specifying evaluation tasks, making things more consistent: https://github.com/neulab/xnmt/issues/260

### DIFF
--- a/examples/standard.yaml
+++ b/examples/standard.yaml
@@ -5,42 +5,47 @@ standard:
     out_file: examples/output/<EXP>.out
     err_file: examples/output/<EXP>.err
     cfg_file: examples/output/<EXP>.yaml
-    eval_metrics: bleu
+  model: !DefaultTranslator
+    glob:
+      dropout: 0.1
+      default_layer_dim: 512
+    src_embedder: !SimpleWordEmbedder {}
+    encoder: !BiLSTMSeqTransducer
+      layers: 1
+    attender: !MlpAttender {}
+    trg_embedder: !SimpleWordEmbedder {}
+    decoder: !MlpSoftmaxDecoder
+      layers: 1
+      mlp_hidden_dim: 512
+      bridge: !CopyBridge {}
+      label_smoothing: 0.1
+    inference: !SimpleInference
+      beam: 5
+      len_norm_type: !PolynomialNormalization
+        apply_during_search: true
+    src_reader: !PlainTextReader {}
+    trg_reader: !PlainTextReader {}
   train: !SimpleTrainingRegimen
     run_for_epochs: 20
     batcher: !SrcBatcher
       batch_size: 32
-    glob:
-      dropout: 0.1
-      default_layer_dim: 512
     restart_trainer: True
     trainer: !AdamTrainer
       alpha: 0.0002
     lr_decay: 0.5
     dev_metrics: bleu
-    corpus_parser: !BilingualCorpusParser
-      src_reader: !PlainTextReader {}
-      trg_reader: !PlainTextReader {}
-      training_corpus: !BilingualTrainingCorpus
-        train_src: examples/data/train.ja
-        train_trg: examples/data/train.en
-        dev_src: examples/data/dev.ja
-        dev_trg: examples/data/dev.en
-    model: !DefaultTranslator
-      src_embedder: !SimpleWordEmbedder {}
-      encoder: !BiLSTMSeqTransducer
-        layers: 1
-      attender: !MlpAttender {}
-      trg_embedder: !SimpleWordEmbedder {}
-      decoder: !MlpSoftmaxDecoder
-        layers: 1
-        mlp_hidden_dim: 512
-        bridge: !CopyBridge {}
-        label_smoothing: 0.1
-  inference: !SimpleInference
-    beam: 5
-    len_norm_type: !PolynomialNormalization
-      apply_during_search: true
-    src_file: examples/data/test.ja
+    src_file: examples/data/train.ja
+    trg_file: examples/data/train.en
+    dev_tasks:
+      - !AccuracyEvalTask
+        eval_metrics: bleu wer
+        src_file: examples/data/dev.ja
+        trg_file: examples/data/dev.en
+      - !LossEvalTask
+        src_file: examples/data/dev.ja
+        trg_file: examples/data/dev.en
   evaluate:
-    ref_file: examples/data/test.en
+    - !AccuracyEvalTask
+      eval_metrics: bleu wer
+      src_file: examples/data/test.ja
+      trg_file: examples/data/test.en


neubig:
OK, I've done an initial refactoring in `test/config/standard.yaml` and things seem to be somewhat working. This has introduced several critical issues:

- [ ] Tests are failing
- [ ] Reporting of development evaluation speed/word is broken

I'll start working on these, starting with the first. The second is not hard code-wise, but I'm still thinking of the most conceptually clean way to do it.

It has also broken some existing functionality:

- [ ] It is no longer possible to automatically calculate the vocabulary size because the text parser is no longer associated with a training corpus. This is somewhat annoying, but also I think it's the right way of doing things considering that we might be doing multi-task learning with multiple corpora. I'm also thinking about the most appropriate way to fix this in a correct way that doesn't add a lot of inconvenience. One option is to assume that we must pre-calculate and pass in a vocabulary file if we want to know the vocabulary size in advance. Another would be to make it possible to pass in a corpus to the `Vocab` class.
- [ ] Shuffling batches from epoch to epoch is broken.
- [ ] Corpus filtering at training time is broken for max sentence length and max number of sentences. I'm not sure if the former functionality is even necessary (it can already be done in pre-processing). The second should probably be fixed.

There are also some non-ideal parts of the configuration file.

- [ ] It is necessary to explicitly reference the model several times in the configuration file.

I'll continue working, but @msperber if you have any comments or could potentially help tackle the last issue of duplicated model references that'd be appreciated!

msperber:
- I agree it's a good idea to remove the automatic vocab size calculation, for the reasons you mentioned and also because it would allow us to get rid of the dependent init parameters that are used only for this purpose and make the serialization code unnecessarily complicated. Initializing the vocab by specifying vocab_file is already implemented, although from_corpus might be more convenient in some cases.
- I think corpus filtering at training time via max sentence length is also useful and should be kept intact because people might have large corpora (e.g. speech data) in which case using the preprocessing feature is a bit annoying because it costs a lot of disk space.
- I'm thinking of adding a !Ref tag that would replace the current reference-sharing mechanism. It would be used somewhat like this: !Ref{path:model} or !Ref{path:train.tasks.0.model.encoder}. Besides requiring less boilerplate, it would allow setting default values such as AccuracyEvalTask.__init__(model=Ref(path="model")). Would that make sense to you?

neubig:
OK, sounds good on 1 and 2. I'll make sure we can keep this functionality then.

For 3., I think this basically sounds like a good idea but am not 100% sure of the logistics. I think it would still be nice to do references by name and not path, perhaps keeping the `_xnmt_id=XXX` syntax and allowing things like `Ref(id=XXX)` as well as `Ref(path=YYY)`. What do you think? If that sounds good, could you try and whip something up? I'm much more familiar with the serializer now than I was before, but I still think that you could do it better.

msperber:
I worked on the serialization a bit and !Ref is working with either a path or a name now, but I haven't finished the part that replaces Ref objects in the default arguments.

I did have an idea for circumventing the problem with the unknown vocab size: How about we add src_vocab_fixed(vocab, task) and trg_vocab_fixed(..) events that can be listened to by embedders etc? These can then simply perform the parameter allocations that depend on the vocab size in the listener rather than in the constructor. That would also make other things more flexible, e.g. remove the need to read the corpus in the constructor, etc.

neubig:
Great! Once you get the first "Ref" part done push it to this branch and I'll try to handle the rest. This is greatly appreciated, and I think it will add a lot of flexibility and reduce complexity when it's done.

For the second part, I'm just a little bit apprehensive as it seems like it might result in added complexity that might not be intuitive. If you think it's worth a try go a head though!

msperber:
Okay, I pushed my changes:
- I ended up mostly rewriting the serialization code. Conceptually it's exactly as before, but I refactored to re-use the tree traversing logic everywhere. Previously there was a lot of duplicated and sometimes inconsistent tree-traversing code that made it difficult to change the serialization behavior. I hope that the code is now much easier to understand and extend. (It's not impossible that there are still bugs).
- Components in the tree can be referenced via a path, and the new Ref object allows reference-sharing. References are either a full path, or a name that corresponds to a _xnmt_id attribute somewhere in the tree.
- Ref objects can be used as default parameters, as is currently done in SimpleTrainingRegimen.
- I updated the standard.yaml example, it's running through until the dev checkpoint
- Note one difference in deserialization behavior: Previously, different parts such as train, eval, .. were deserialized separately. Now everything is deserialized jointly to make referencing across the whole experiment possible.
- The new code does not support dependent_init_params any longer. Actually, I have some hope that it would be possible to use default Ref arguments to also share the vocab across components. If this or other simple solutions don't seem possible, I think I would prefer the event-based approach over the dependent_init_params approach because it's conceptually simpler.

neubig:
Thanks! I'll take a look.

neubig:
When you said `standard.yaml` is working did you mean `python xnmt/xnmt_run_experiment.py test/config/standard.yaml` should work? I got the following error:

    Traceback (most recent call last):
     File "xnmt/xnmt_run_experiments.py", line 133, in <module>
          sys.exit(main())   
     File "xnmt/xnmt_run_experiments.py", line 98, in main
          exp_args["train"].run_training()   
     File "/Users/neubig/work/xnmt/xnmt/training_regimen.py", line 125, in run_training
          loss = self.training_step(src, trg)   
     File "/Users/neubig/work/xnmt/xnmt/training_task.py", line 273, in training_step
          standard_loss = self.model.calc_loss(src, trg, self.loss_calculator)   
     File "/Users/neubig/work/xnmt/xnmt/translator.py", line 119, in calc_loss
          dec_state = self.decoder.initial_state(self.encoder.get_final_states(), self.trg_embedder.embed(ss))   
     File "/Users/neubig/work/xnmt/xnmt/decoder.py", line 96, in initial_state
          rnn_state = rnn_state.set_s(self.bridge.decoder_init(enc_final_states))   
     File "/Users/neubig/work/xnmt/xnmt/decoder.py", line 189, in decoder_init
          raise RuntimeError("CopyBridge requires enc_dim == dec_dim, but got %s and %s" % (enc_final_states[0].main_expr().dim()[0][0], self.dec_dim))   
   RuntimeError   :    CopyBridge requires enc_dim == dec_dim, but got 64 and 512

msperber:
No, sorry, I was referring to examples/standard.yaml

neubig:
Oh, OK... It looks like the `standard.yaml` file is largely in the old format though and doesn't include critical parts like evaluation. `test/config/standard.yaml` is in the refactored format. Could you take a look and see if you could get this to work with your new code? After I have one working example I can do the rest.

msperber:
Got it, test/config/standard.yaml should be working now!

neubig:
Maybe you forgot to push? I don't see any updates.

msperber:
sorry, here it is!

neubig:
Thanks a lot! I fixed `eval_tasks` to also support the top-level model by default.
I started doing the other ones, but got a bit confused on the `component_sharing.yaml` file, specifically the `<< *exp1` syntax? Would it be possible to have you do just this one as well? I can probably do all the others once I know that one.

msperber:
Okay, it's done! I used << mainly to copy over some arguments but then overwrite some other parts (see [here](http://yaml.readthedocs.io/en/latest/example.html) for some brief documentation)

neubig:
OK, I've fixed some things and identified some problems. @msperber, the first two below are related to the serializer, and it'd be nice if you could take a look. The remaining ones I'll try to take a look at.

- [ ] `prior_segmenting.yaml`, `segmenting.yaml`: I'm getting a problem with `xnmt.segmenting_encoder.SegmentingSeqTransducer` getting `input_dim` being specified as an input argument despite the fact that it seems like it's not passed in in the yaml file.
- [ ] `encoders.yaml`: This seems to be failing because `model` is not instantiated in the default parameters. I'm actually not sure why this is a problem though...
- [ ] `test_reload.yaml`, `test_reload_exc.yaml`: I think the reload capability is probably broken and needs to be fixed. 
- [ ] `multi_task.yaml`, `test_random_search_test_params.yaml`: I think these are failing because the softmax is larger than the vocabulary, which we can fix.
- [ ] `transformer.yaml`: `TransformerTranslator` needs to be updated or deleted and replaced with `GeneralTranslator`. I strongly prefer the second if possible.

Also, it seems that it's now necessary to specify `lstm_dim` to `mlp_softmax_decoder`. Can we avoid this?

msperber:
The problem in encoders.yaml is that the "model" must be specified before "train" (or in general, things referred to by the default Ref argument must be already initialized). The cleanest way to enforce this would probably be to make the top-level Serializable instead of a dict, like so:
```
exp1: !Experiment
  model:
  train:
  ...
```
In this case, the arguments will be initialized according to the order they appear in ```Experiment.__init__()```. It does make things slightly more verbose, but I can't think of a good different way, so if it sounds okay I will implement this?

msperber:
Nevermind my previous comment. I changed the code to properly resolve references in forward and backward direction. I also found a workaround for the issue that caused having to specify lstm_dim, although if necessary we can revisit this later and make it more pretty.

neubig:
I fixed the reloading capability.

@msperber I thought a little bit more about the vocab size problem, and I realized that I had made a silly mental error. I was thinking that as long as we set the vocab size appropriately at initialization time then our problems would be solved, but actually the problem of sharing these with the appropriate word embedders still exists. I think your idea about how to fix the problem with events is reasonable, so could you take a look at doing that? I'll try to work on the other issues cited at the top of the thread.

msperber:
I gave it a shot. I realized that sending events is probably not the appropriate way of handling the vocabs because there are too many dependencies, so I just tried to handle it in a very straightforward way. Let me know if you have any concerns? Also, the model saving seems to not be working for the component_sharing example currently.

msperber:
Hmm, this is turning out to be more difficult than expected: the references, together with the fact that we're only saving parts of the YAML tree, make it very hard to ensure we are loading things in the same order from the original config and the saved model file. It might be possible to ensure with some effort, but I'm starting to think it would be better to introduce named DyNet parameters. Basically, any component that requests yaml_context.dynet_parameter_col will be given some name (such as <class_name>-<random-id>) that will be used to name the parameters, and the same name will be used when loading from a pretrained model file. I won't have time for this until next week, so if you have time @neubig please feel free to look at that. With that working, the last piece to make model saving/loading work would be to make sure YamlSerializer.dump() produces Ref objects instead of YAML anchors.

neubig:
Hmm, I'd prefer avoiding making things even more complicated, and I'm not a huge fan of saving a `vocab' object within the `Embedder` and `MlpSoftmaxDecoder` classes. How about we go back to my original plan where the `Vocab` object needs to know its vocabulary at initialization time, either by passing in a vocabulary list, or by passing in a corpus, which the `Vocab` object then reads and uses to determine its vocabulary. We would still need to have a reference to the `Vocab` object, but we wouldn't need to do any initialization of parameters post-facto.

Does this sound reasonable and also sound like it would fix the problem? If so, I can take a try at it.

msperber:
Sure, yeah I think that's a good plan. It might be difficult to specify default Ref arguments for the vocabs because e.g. a SimpleWordEmbedder doesn't know if it's being used with src_vocab or trg_vocab, so the config files would get a little more verbose, but I also think it would make things simpler overall.

Unfortunately it won't solve the inconsistent initialization order problem, which is caused by the Ref system and by having to serialize "model" into "train.model" because only "train" is serialized. But I thought about it a bit more and think when initializing things for the first time, I would attach an _init_order integer to each object that gets written out to the model file and can be used to recover the same order when loading a saved model. That would only affect the serializer code and seems simpler than named DyNet parameters if I'm not missing anything.

neubig:
OK, I'll take a look at doing this soonish. Two questions:

1. Is there a way that we could make things aware of their path in the yaml file when calling their constructor? For example this would allow us to specify defaults like the following:

    class VocabEmbedder(..., vocab=None):
      ...
      if vocab == None:
        if "src_embedder" in find_my_yaml_path():
          vocab = Ref(Path("..", "src_reader", "vocab"))
        elif "trg_embedder" in find_my_yaml_path(): 
          vocab = Ref(Path("..", "trg_reader", "vocab"))
      if vocab == None: # If still not found
        raise ValueError("No valid vocabulary object found")

2. For the `_init_order` integer, I'm happy to take a look at this, but do you have an idea where would be the best place to do this?

msperber:
Regarding 1: That should be easy be inserting something like this under https://github.com/neulab/xnmt/blob/44dcea7801e7a7d9dbe8cf4d1054d4357355014e/xnmt/serialize/serializer.py#L127 
```
if "yaml_path" in init_args: obj.init_params["yaml_path"] = path
```
Then, whenever ```__init__``` accepts a yaml_path argument, the path would be passed automatically. However, the Ref objects are resolved *before* initializers are called, so we can't create new Ref objects inside the initializer. Instead we might end up with something like:
```
class VocabEmbedder(..., vocab=None, src_vocab=Ref(["..","src_reader","vocab"]), trg_vocab=Ref(["..","trg_reader","vocab"])):
  ...
  if vocab == None:
    if "src_embedder" in find_my_yaml_path():
      vocab = src_vocab
    elif "trg_embedder" in find_my_yaml_path(): 
      vocab = trg_vocab
  if vocab == None: # If still not found
    raise ValueError("No valid vocabulary object found")
```

Note also that currently no relative paths are supported, but that shouldn't be too difficult to do.

Regarding 2, it might look something like this:
- in YamlSerializer.init_components_bottom_up(), keep a counter during the for loop and add  to each initialized component's serialize_params dict a corresponding field using ```set_descendant(root, path.add("serialize_params").add("_init_order"), ctr)```
- if ```_init_order``` fields are present, make sure that ```tree_tools.traverse_tree_deep_once()``` returns nodes in that order, e.g. by first producing all items and then sorting according to this field
- also, currently when saving a model this results in shared objects to become YAML anchors, though they would need to become Ref objects instead. Doing this is not super easy, it might be better instead to make sure tree_tools.traverse_tree_deep_once() returns the same path whenever ```_init_order``` is the same, that should be enough to result in properly shared objects due to YamlSerializer.init_component()'s caching property.
- add ```_init_order``` to tree_tools.reserved_arg_names, and make sure the field is properly ignored by the tree_tools helpers.

neubig:
Thanks! I'll try to take a look today.

neubig:
OK, I think I basically tackled 1. above. There are still 6 varieties of failing tests, some regarding serialization itself and some just miscellaneous errors. I'll try to tackle the miscellaneous ones myself, and perhaps leave the serializer-related ones for you to take a look at when you're back @msperber .

msperber:
I implemented writing out of Ref objects when saving a model, and improved a few other places. The initialization order issues disappeared also, even without making the order explicit. I think this should be okay for now, but I'm not 100% if we might find corner cases in the future that would require revisiting this. There are a few remaining mixed issues, would you take a look at those @neubig ?

msperber:
The remaining 2 broken tests still seem related to model loading, I'll take a look tomorrow.

neubig:
OK, great, thanks! That part of the code you're more familiar with than I am.

msperber:
I reworked the model saving and re-loading process. As of the new approach, the whole experiment is serialized to YAML (instead of just the TrainingRegimen), and xnmt_run_experiment is in charge of this. This makes saving/loading much more consistent, fixes the initialization order issues, and integrates more nicely with things like multiple models. This also made the cfg_file parameter obsolete.

Loading a pretrained model and overwriting parts of it is now also more intuitive and less verbose, especially when changing only a small part of a model (see load_model.yaml example).

Tests are passing, but there are still some TODOs in SimpleTrainingTask.__init__() which I hope would be possible to fix before merging? I saw the comment about not supporting Refs for the preproc part, but don't think it can be fixed with the current design, unfortunately. There are also TODOs left in the config files, let me know if I should take a look at those.

msperber:
I would have preferred the syntax you mentioned also, but "." is not allowed in dictionary keys (otherwise resolving references would become ambiguous). We could think of escaping etc. though if you would like.

And there is still one difference between the user-specified YAML file and the saved YAML file: the latter contains only one experiment. So the overwrite part always refers to the model loaded within the same experiment. Currently, the <EXP> is resolved when parsing the YAML file, but then it's actually not written back into the saved YAML file because of some pop() operations in xnmt_run_experiments. So when loading a pretrained model without overwriting the model file, it will actually just write the new model to the default model location.

If we wanted to do this in a cleaner fashion, one way would be to use python string formatting syntax such as modelfile_str="/some/path/{exp_name}.mod" and leave that string as is, and then applying modelfile_str.format(**some_args) only when actually saving the model, but writing the unparsed string back into the saved model file.

neubig:
Thanks for the clarification. For the first part (syntax), I understand the reason now, so let's just go with this as-is for now and think about potentially changing later.

For the second part, got it, that makes sense. It might be nice to document this somewhere if it isn't documented already.

For the third part, I'd prefer this if it's not too much work.

msperber:
Ok, done!

msperber:
I fixed one of the TODOs in the code that required changing the serialization. All other TODOs (except the preproc one) should be possible without any more adjustments of the serializer.

Looking through the code, I had 2 more ideas:
- The training tasks have a ```name``` that shows up in the logs so we can distinguish between them. Since we might have several evaluation tasks, they could be named as well to make log files easier to read.
- the YAML serializer is currently invoked twice: first for preproc (a subobject of experiment), then for the whole experiment. This means there can be no reference sharing, and it's also a bit dangerous because the preproc objects will be initialized twice, which could be expensive depending on what they do during initialization, or could even lead to bugs due to side effects. A possible way to fix this would be:
  - wrap the preproc modules in one ```!PreprocRunner``` that calls ```xnmt_preproc()``` in its initializer.
  - make the top level entry an ```!Experiment``` object. By putting ```preproc``` at the beginning of ```__init__``` arguments, we can make sure the preproc runner is invoked before initializing the training code.
  - the Experiment object will have a run() method that is invoked by xnmt_run_experiments. The code inside the experiment loop except for logging and deserialization is moved into Experiment.run().
  - this might require changing the way the yaml_context is handled. I think with the new Ref system, we could just do something like ```Embedder.__init__(..,yaml_context=Ref(Path("global")), ..)``` and remove the special handling of it altogether.

Not sure if these should be part of this PR or a future one though.

neubig:
I think the first of these is easy. For the second, yes, this is quite ugly. I like your first proposed situation of wrapping it in `!PreprocRunner`. I don't have time to do either of these today, but if you can it'd be appreciated!

msperber:
I took a look but am not very familiar with how xnmt_preproc works and am afraid I might break things. But if you could add a PreprocRunner that handles everything in its initializer and doesn't make calls to the YamlSerializer, then I can handle the rest (implementing the Experiment top-level object), that will be easy for me.

msperber:
I implemented the ```!Experiment``` as mentioned above which will take care of initializing things in the proper order and also modularizes things a bit further. A few tests are failing, including the preproc test (which still needs implementation of the PreprocRunner), and a few others which I will take a look at soon.

neubig:
@msperber Thanks! I might have a little time to work on this today, but won't start if things are still in flux. Tell me when things have settled down and it'd be worth me looking.

msperber:
I won’t be changing anything more this week, so feel free to go ahead!

neubig:
I wasn't able to get to this last week, but I'll check later today when you're done with what you're doing now!

msperber:
No problem! I did some cleanup and fixes and think that things are now in a consistent state and could be merged from my side. There are still some details that can be improved (not having to specify the default inference, naming for eval tasks, ..), but they are not super important and it might be better to merge this first so that people can add new features without risking merge conflicts.

neubig:
Thanks! It looks like everything is working now. I'll be doing some stress tests, and fix anything I find at that point.

Issue #269, opened 2018-01-17
Target sequence truncated in report (translator.py)


pierregodard:
When force decoding in a seq2seq model (config file similar to forced.yaml), it appears that the following line (https://github.com/neulab/xnmt/blob/master/xnmt/translator.py#L130)
```
trg_words = [self.trg_vocab[w] for w in output_actions[1:]]
```
truncates the target sequence from the first unit. Correcting to `output_actions[0:]` solves my problem (I am dumping attention matrices in a plain text format), but I wonder if there is another report scenario in which the truncation is needed?

PR #270, opened 2018-01-18
Add plain text report capability to DefaultTranslator

pierregodard:
Add `report_type: txt` option in the yaml config file to report DefaultTranslator's attention matrices in a plain text format (lines indexed by the target sequence, and columns indexed by the source sequence). 

See [#269](https://github.com/neulab/xnmt/issues/269) however, as this could break something somewhere else.

Also `png` option seems to be coupled with `html` option which isn't natural (I don't think you can get a png report without an html one). I could fix this unless I miss something.


philip30:
Hey Pierre, I took a look at it. 1 comment I want to make is that you maybe can merge the txt report into file report. They function the same way I think. I actually already made the same thing but in the different branch. But I think I can merge mine into yours later.

pierregodard:
Thank you Philip for taking a look at it! The `on_file_report` method in `SegmentingSeqTransducer` is the only implementation of file report that I see, and it seems it was meant to report for segmented sequences, not attention matrices. But you're right, if both are plain text reports they should rely on the same report type. Should I keep 'txt' instead of 'file' in merged method names, to indicate a plain text report? It feels like 'html' or 'png' are *files* too, no?

philip30:
Hi Pierre, yes you are correct. Since there are many interpretations of the file, I actually leave the file report for general purpose. e.g. for segmentation it is the segmentation, and translator for attention matrix. Maybe txt is better naming, but I just want to avoid adding 1 more type of xnmt_event. Maybe we can refactor it later to rename from file to txt.

philip30:
Also, I think it is OK to generate the png alongside the html. The png report is a very rare case. So for the sake of simplicity of the code, I think it is OK to generate both and you can just copy all the generated png if that is what you desired.

pierregodard:
Thank you @philip30. Got it for the png/html rationale.

However, as I'm trying to merge the `file` and `txt` reporting options, I'm noticing something surprising in the way the reporting is handled for `html` and for `file`.

In `reports.py` we have two different logics apparently (notwithstanding the `context`/`assign` difference):
```
@register_xnmt_event_assign
  def html_report(self, context=None):
    raise NotImplementedError()
```
and
```
@register_xnmt_event
  def file_report(self):
    pass
```
Then in `translator.py`, we can find the following implementation:
```
@register_xnmt_event_assign
  def html_report(self, context=None):
  ...
```
while in `segmenting_encoder.py` we find the "on_event" handler form:
```
@handle_xnmt_event
  def on_file_report(self):
  ...
```

I followed the former for plain text report but I now see that the later (in `segmenting_encoder`) seems more consistent with the way events are handled in xnmt (actually the other method doesn't seem to even rely on events). The empty method with only `pass` instruction doesn't feel totally natural though, as it seems that its only role is to trigger the event. I hope I am not missing something obvious. 

Can you help me understand the best way to merge `txt` and `file` reports? (with the "file" logic or the "html" logic). And maybe subsequently the best way to harmonize both already existing reporting capabilities (`html` and `file`)?

philip30:
I think @msperber is more appropriate in answering this question because I am not very familiar with the handling of the assign_event...

I actually had already done that in the separate branch, you can take a look at https://github.com/neulab/xnmt/blob/segmentation_experiment/xnmt/translator.py#L230. Please tell me if this helps you? :)

msperber:
Yes it's totally possible that the reporting-related events are inconsistent or even obsolete or broken. In general, events are meant to be triggered with methods annotated via ```@register_xnmt_event*``` , triggering all methods named ```on_<trigger-name>``` annotated with ```@handle_xnmt_event```. The trigger can, but doesn't have to, do things, so in the second case we would need the pass statement.

I'm not very familiar with how the reporting works or should work, but please feel free to adjust or drop usage of events as you find appropriate.

pierregodard:
Thank you both very much for taking a look at this! I initially set up to remove the usage of events here as I didn't see why it was necessary and in order to simplify the code. This worked fine for `translator.py`. But I then had to change my mind since I discovered that `segmenting_encoder.py` uses the `context` argument (see [segmenting_encoder.py:250](https://github.com/neulab/xnmt/blob/master/xnmt/segmenting_encoder.py#L250)) that comes with the event mechanism, and I didn't want to break Philip's reporting.

Going the other way around (see my latest commits) makes the `segmenting.yaml` test fail though, as the inference calls the html reporting from `translator.py`, which in turn also triggers the `on_html_report` from `segmenting_encoder.py`, but with a context argument set to `None`.

I am a bit confused now. It seems that implementing `on_<trigger_name>` for reporting in different places will then result in all these handlers being triggered by one reporting event and I don't think it's the desired behavior for reporting, or is it? 

@philip30 : I could push a 'no event' version of reporting but I'm not clear then how it would affect your usage of the context in [segmenting_encoder.py:250](https://github.com/neulab/xnmt/blob/master/xnmt/segmenting_encoder.py#L250).

philip30:
Hi @pierregodard, I think there was a reason why it was made that way, so it'd best not to change the type of the event. Let's assume that it was thoughtfully made that way. For the html reporting, it needs the assign type event because the recursion is made this way:

```python
context = self.report(None)
for child in self.children:
    context = child.report(context)
return context
```
So basically, the html (the context) is constructed from the parent (translator) and the children (encoders, decoders, attenders, etc). I think you can't make the reporting without the event.

pierregodard:
Got it @philip30. And I now understand why the `segmenting.yaml` test was failing. 

It initially seemed reasonable to change, as I did, 

```python
@register_xnmt_event_assign
def html_report(self, context=None):
  raise NotImplementedError()
```
to
```python
@register_xnmt_event_assign
def html_report(self, context=None):
  pass
```

and in `translator.py`
```python
@register_xnmt_event_assign
def html_report(self, context=None):
  ...
```
to
```python
@handle_xnmt_event
def on_html_report(self, context=None):
  ...
```
to be consistent with the `file_report` logic, but I now understand that `html_report` in `translator.py` was in fact intended to override the `events.py` method, in other words to be a *trigger* and not to be an event *handler*. 

I am reverting my last two commits.

philip30:
It looks great! You can merge anytime you think is appropriate.

pierregodard:
Thank you @philip30. Now I have a branch conflict with recent [#267](https://github.com/neulab/xnmt/pull/267) merge. I thought I just resolved that conflict, and I have a [merge](https://github.com/neulab/xnmt/pull/270/commits/2a878b72e8afc24d9c349262f9bdd2d9909fdbba) showing above but still a conflict! Will try to understand but maybe you'll see instantaneously the problem... Thank you for all your help.

philip30:
Sorry @pierregodard, I am also not very familiar with the most recent pull request. Maybe @neubig can help. BTW, can @neubig answer the question "Why did we truncate the first output action?" I am also not really sure why.

philip30:
I think I start to understand what that was. The first output action is the start symbol I guess "<s>", that is why it is deleted. I think the appropriate way is to delete the first column from our attention matrix.

pierregodard:
thank you @philip30. I did the 'resolve conflicts' and merged again, the same way I did before AFAIK, but this time it worked, no more branch conflict.

yes, it would be great if @neubig could tell us if the truncation mentioned in [#269](https://github.com/neulab/xnmt/issues/269) was indeed a bug and should be corrected as I did here, or if it was serving another purpose and I'm breaking something else now.

philip30:
I just want to let you know that the current way of storing the attention matrix in the translator is broken and wrong. When you use beam search, all the attention vectors are stored linearly in the attender! (This line does not make sense then: https://github.com/neulab/xnmt/blob/master/xnmt/translator.py#L135) This is wrong and needs to be fixed. I am fixing it separately on my diverged branch, but you might also want to fix this. @pierregodard @neubig

pierregodard:
Thank you @philip30 for the insight. I did notice that and was puzzled about it, but since my use case is to dump attention matrices produced during forced decoding, which assumes a beam size of 1, it wasn't an issue for me here. 

Let's see what @neubig  says about the truncation question we had, but in any case do you think we should address this last issue you mention before I merge this? My understanding of the current implementation of the beam search is not totally clear at this point, but I could look into it.

philip30:
The first element is the <s> sentence and I think we can either drop it + drop the first attention column or just include them altogether. But in most cases, I don't see the <s>, so I think we can add <s>. There are several cases also where the output of the first word is truncated.

pierregodard:
@philip30: are you talking about the source or target side here? (or both?) You mention dropping the first attention column so that would mean the source side but I suspect you might mean the target side, as in xnmt the convention seems to be the other way around for attention matrices (see my comment [here](https://github.com/neulab/xnmt/pull/270/commits/297c054173cde653363957c8aa138c0a96d6bb54#diff-fd01cb9452fd023aa34552ca193efb5cR215)). And I was also confused about the other cases of truncation you mentioned at the end of your last comment, maybe you can clarify?

I don't find usage of the start symbol like I do for the end symbol (e.g. [input.py:147](https://github.com/neulab/xnmt/blob/master/xnmt/input.py#L147)). At least in my experiments with forced decoding, the end symbol `</s>` is added to the output sequence, but not the start symbol `<s>`.  So the truncation practically cuts the first 'real' symbol of the (forced) output sequence.

philip30:
Hi Pierre, the <s> was meant to be generated not read, but I didn't see the translator generated that either. BTW, my conclusion is not dropping anything and make a sanity check that the length of the attention_matrix.shape == (len(src), len(trg))

I have a separate working implementation about this actually on segmenation_experiment branch: https://github.com/neulab/xnmt/blob/segmentation_experiment/xnmt/translator.py#L233
You don't have to follow mine, because I will merge far later. You can merge earlier and we can think a way to resolve them in the future. 

The output was truncated because the first word is dropped. And I fixed it also in my branch.

pierregodard:
Thank you Philip for the clarification. I just added your sanity check. I could merge this when Graham confirms that the removal of the truncation doesn't break something else. 

(before I merge, however, you need to know that I chose the convention to index lines with the target units, and columns with the source units, so it's summing to 1.0 line-wise. I mentioned it above but wanted to make sure you saw it.)

xnmt/translator.py
```diff
@@ -131,7 +132,9 @@ def generate(self, src, idx, src_mask=None, forced_trg_ids=None):
       # In case of reporting
       if self.report_path is not None:
         src_words = [self.reporting_src_vocab[w] for w in sents]
-        trg_words = [self.trg_vocab[w] for w in output_actions[1:]]
+        # next line seemed to truncate first trg elt in the seq (in forced decoding mode at least)
+        # `output_actions[1:]` replaced by `output_actions[0:]`
+        trg_words = [self.trg_vocab[w] for w in output_actions[0:]]
```
neubig: This should not be `[0:]`, you can just remove this.

pierregodard: Thank you very much for taking a look. But if I put back `output_actions[1:]`, then the target is truncated from its first unit, and the attention matrix's shape doesn't match (len(src), len(trg)) anymore. Is that what you meant?

neubig: No, I meant remove `[0:]`, leaving only `output_actions`. `[0:]` is a noop and just makes the code more complicated.

pierregodard: Oh sorry, yes. So silly of me. Done.

neubig:
Thanks!

philip30:
BTW this will break the decoding if we decide to decode with beam search. The added sanity check will not let the reporting happens. I hope a follow up pull request fixing the attender can be done.

neubig:
@philip30 Wait, how serious is this? Will all decoding break with beam search? Only when reporting is turned on? Or will the reports just be wrong? If this breaks decoding with beam search, I didn't realize this when merging, but we should fix it ASAP.

philip30:
@neubig It will break when decoding is performed with beam search if reporting is turned on. As long as you are doing the normal decoding, that wouldn't be a problem.

neubig:
Ok, good. Thanks for the clarification

pierregodard:
Maybe added the sanity check was unwise then. @philip30 you seem to have already changed in another branch the way attention vectors are stored in the attender, but if you want to give me a pointer on what you did, I could try to do the follow up pull request you're talking about. If I understand correctly, with beam search we should only keep the attention matrix corresponding to the last best hypothesis, right?

(Also, as I use the plain text report in forced decoding, it is already asserted in the code that the beam size is 1, but you're right, nothing prevents anyone to turn on file reporting while performing beam search.)

philip30:
@pierregodard I think it is more important to do the things correctly than working. A decoding with a beam search is a very common case in text-to-text machine translation. And we were wondering why the results are incorrect. After a few months, I just discover this. So the sanity check is necessary, I guess.

What I did was specified in this commit:
- https://github.com/neulab/xnmt/commit/f222a6c4cca987a1453ce83def78385db7f3657b
- https://github.com/neulab/xnmt/commit/1f1894e639cc55591db92e0c868b742c2754353d [only the search strategy and the attender]

philip30:
Yes, we should add the attention matrix to each of the beam hypothesis.

pierregodard:
Thank you Philip. I'm obviously less familiar than you with all this, and I will need a bit more time, so if you want to fix it in the master branch in the next 24hrs, it's probably safer that you do it. But otherwise and if it's useful for you to able to continue focussing on your branch, I'm happy to look into those two commits you linked, and make a separate pull request to fix the way attention matrices are stored during beam search. Let me know!

PR #274, opened 2018-02-03
Fixes and simplified config files

msperber:
This fixes a few issues that became apparent after the recent refactoring. Most significantly, config files are made more compact and easier to write thanks to better error handling and a mechanism for Serializable default parameters.

Details:
* introduced a consistent way of specifying Serializable objects as default arguments in ```__init__()```. Previously, the problem was that something like ```__init__(…, inference=SimpleInference())``` would often fail because the default object would be initialized prematurely and skip referencing / param sharing mechanisms etc. This is now solved via the bare() wrapper that prevents objects from being instantiated until they are actually parsed by the YAML deserializer. It’s used as ```__init__(…, inference=bare(SimpleInference,param1=v1,..))```. I added defaults to most of DefaultTranslator’s arguments, allowing us to skip a lot of arguments in the config files, see the updated standard.yaml.
* better error handling for deserializer; this gets rids of some obscure error messages and also allows simplifying config files further.
* less confusing load/overwrite syntax: overwriting options are now a dictionary with keys “path” and “val”, instead of a list.
* evaluators now have a desc field that will appear in the log output
* rename xnmt_global -> exp_global. Sorry for yet another rename, but I believe this communicates more clearly that this object is to be seen as global to the experiment, not XNMT in general.
* add max_src_len / max_trg_len to LossEvalTask
* fix writing out of eval results in log file (currently getting lost)
* fix xnmt_evaluate.py commandline usage
* fix crash when curly brackets appear in vocab
* add missing start_sent event trigger to TransformerTranslator
* skip cython tests if cython unavailable


neubig:
Thanks!

PR #276, opened 2018-02-07
Logging

msperber:
This PR introduces a Python logger to replace most print() outputs. I tried to follow the official [Python conventions](https://docs.python.org/3/howto/logging.html#when-to-use-logging) that suggest using print() only for necessary outputs without which the program execution wouldn't be complete (e.g. final results), and logger.debug(), info(), etc. otherwise. The main advantage is that this allows controlling the verbosity, and make use of other nice features of Python logging. A slight change in behavior is that we no longer have an err_file, instead everything is logged to out_file.

I also added a structured logger that dumps Python dictionaries into a .yaml log file. I found this very useful for dumping things that I want to analyze during training or decoding, but that are too bulky to output using print statements.

I added some comments to programming_style.rst on how these should be used. The guidelines are followed everywhere except in the segmenting-related code which I would leave up to @philip30 .


msperber:
Yes, definitely!

philip30:
Maybe we should also change the out_file into log_file for consistency?

msperber:
Sure :)

neubig:
Thanks!

PR #277, opened 2018-02-08
Initialization

msperber:
- Make parameter initialization more consistent by initializing all biases to 0. This is the proper way to do things, but currently is happening only with LSTM biases (except forget gate).
- add a ```glorot_gain``` parameter to customize the scale of the initialized values. It can either be set individually for each component or globally using ExpGlobal. This is somewhat of a usability-vs-flexibility compromise: we could also make the complete initializer configurable, but that makes model specification more complicated and adjusting the scale is usually more important than other properties of the initial parameter distribution.


msperber:
@neubig @philip30 let me know if you'd prefer the more flexible approach of specifying the complete initialization strategy rather than just the glorot_gain. It might make it easier to reproduce results from the literature, and, for example, would allow moving some of the specialized transformer-code into the config files. If glorot_gain seems enough, this would be ready to be merged. I've been using it extensively in the meantime. Initializing the biases to 0 didn't make a big difference, but is conceptually cleaner and removes one item from the bugchecking list.

philip30:
How about standardizing this by making the add_param_with_default_init and add_bias_with_default_init or something like that to the exp_global or init module? I meant, we encapsulate the param_collections.add_parameter with some method. I think this can encourage other developers to use this initialization.

Other than that, I think this is LGTM.

neubig:
The naming `glorot_gain` will be really hard for a non-expert to understand I think. Maybe `init_scale` or something like that?

And I understand the tradeoff here, but I think this is something that should definitely be configurable. One goal of xnmt is to be able to more-or-less reproduce previous research by just modifying the configuration, and initialization is an important factor in this. One option would be to name the option `param_init` or something like that, then if it's `None` then use the defaults, if it's a float, then use it in the same way as `glorot_gain` right now, and if it's a class, use it as a an initializer. We could add this logic to a separate `param_init.py` file perhaps to prevent code duplication.

I also like Philip's idea of having a global parameter that can adjust the default parameter init if desired.

msperber:
That seems reasonable, I'll take a look soon!

msperber:
This should be ready: I added wrappers for the DyNet initializers so they can be specified in the config files. The defaults are set in ExpGlobal: ZeroInitializer for bias parameters, and GlorotInitializer otherwise.

Also fixed a minor bug due to the 4 LSTM matrices that are merged into one parameter matrix, which lets them appear to the Glorot Initializer as having a higher dimensionality than if they were specified separately.

PTAL.

neubig:
Looks good, thanks!

PR #278, opened 2018-02-14
Better ref resolution, fix speech reporting

msperber:
Previously, the reference resolution when saving a model went breadth-first, which caused the model to be specified inside the evaluators and then referenced elsewhere. This made it difficult to load/overwrite certain parts of the model because references are not followed when overwriting. This PR changes to a more natural resolution order, which is probably still not perfect in all situations, but better than before. In the future, we might try following references when doing load/overwrite, and/or keeping track of the reference hierarchy as specified in the original config file.

Also included is an unrelated quick fix for speech reporting.


msperber:
This PR probably doesn't need a review and I have tested/used it extensively in the meantime, so I'll go ahead and merge.

Issue #279, opened 2018-02-26
Example model is not verbose enough


neubig:
Currently in `standard.yaml`, etc., the model file is now very very concise to the point that it is not necessarily a good example because it doesn't demonstrate any of the modifiable parameters. I think it'd be nice if we at least kept around the main elements of the MT system in the yaml file.

mbollmann:
About the example files -- I thought many of them were very useful, but the main things I was missing (but have since figured out) were:

- Beam search decoding!  None of the example files show how to use it, though it is a very important feature, I would think.
- Early stopping.  The `08_multi_task.yaml` mentions early stopping (and learning rate decay) in the comments, but it isn't shown anywhere.

I realize the examples & docs are far from complete, but maybe this is useful for deciding what to add next!

msperber:
An example of loading a model for the purpose of evaluation should also be added.

neubig:
Fixed by #305

PR #281, opened 2018-03-01
fixed and improved examples

msperber:
- fix and reorganize the examples. The idea is to have examples/ serve as kind of a tutorial, and later add a recipes/ directory with configs for tested, state-of-the-art models.
  - config files now begin with 01_, 02_, .. to help people know where to start
  - standard is split into a new standard example with more options specified (addressing #279), and a minimal example.
  - added some more introductory comments to some config files
  - fixed preproc example
  - moved some failing example to the test configs, marked with @unittest.expectedFailure
- dropout default is changed from 0.0 to 0.3, similar to OpenNMT


philip30:
Actually I can't run the test file when there is a line checking there is a setting or not. Do you know the problem? It said that setting is not a recognized argument.

msperber:
Hm could you tell me the exact command you used?

philip30:
I tried both 

```
python3 test/test_run.py
python3 test/test_run.py --settings=settings.unittest
```

Both giving me the same error:
```
usage: test_run.py [-h] [-v] [-q] [--locals] [-f] [-c] [-b]
                   [tests [tests ...]]
test_run.py: error: unrecognized arguments: --settings=settings.unittes
```

msperber:
Ok I see. It works for me like this:
```
python -m unittest test.test_run
```
Thanks for catching it, and I also updated the docs accordingly.

philip30:
Not really sure that I like the way we execute the 'module' rather than a script. Executing a module is a lot trickier I think. Can we preserve the functionality of the executable scripts with the settings?

msperber:
I looked into it and it seems the issue is not script vs module, but that unittests should generally be run via 'python -m unittest', as indicated in the Python documentation: https://docs.python.org/3/library/unittest.html#command-line-interface

But you can also use ``python -m unittest test/test_run.py`` instead.

neubig:
Thanks! This is nice.

Issue #178, opened 2017-08-30
Is Minimum Risk Training needed?


philip30:
I wonder if this will be a good feature?

I also open a discussion about a good design for this training.

neubig:
Yes, this, and other training algorithms such as REINFORCE would be greatly appreciated!

PR #284, opened 2018-03-02
Implementing Minimum Risk Training

philip30:
This is the start of the implementation of minimum risk training. This should not be ready for review yet.


philip30:
This should be ready for review. I want to run a small experiment to sanity-check my code, but before that can @neubig help to see the min risk code if there is an obvious mistake? I also refactored the reinforce code and move the sample method to the model. So this loss is more general and can be used for other model (not only for translator).

neubig:
Actually, I've been meaning to change this for a while: could you change sampling so it's a class `search_strategy.py`, along with beam search and other things? This will allow us to do things like choose between beam search and sampling when we perform minimum risk training, or sample hypotheses at test time instead of searching for the best one. I'll take a look at the other parts in a bit.

philip30:
Yea I think that's totally reasonable, I will do that!

philip30:
I just want to confirm that I successfully increase my BTEC experiment with 50,000 head pairs of training corpus using SGD(lr=0.05). 

0.5096 -> 0.5119

I just wonder if I should continue with the refactoring of the losses? That is kind of different issue, I guess because I need to change the beam search, force decoding, and sampling code. Maybe it is better to make a separate PR for that.

neubig:
This is implementing `sample_one` into the `translator` file, which I think is not ideal design, so given that this PR is not super-high priority (it's nice, but no-one is being blocked by it), I think we should aim to have good design before merging to master.

neubig:
Any update on this? I think it's pretty close to being ready to merge.

philip30:
Sorry, this PR become buried. I'll be working on it soon!

philip30:
This should be ready for review. Although not all the components are finished. These are the changes summary.

1. Refactorsearch strategy to be the part of inference. Inference will not guess the search strategy from args, but will instantiate the requested search_strategy.
2. Allow generator to have the "search_strategy" this search strategy is used during training that requires model to do decoding in the middle. Currently, there are two types of decoding supported: Greedy and Sampling. BeamSearch is quite tricky as input is batched, and it's currently hard to implement the batched beam search decoding / work around with every single sample. (Requires us to do dy.pick_batch_elem one by one for each encoded state)
3. Implemented MinRiskLoss and ReinforceLoss that used the generator search strategy to generate hypotheses and score it during training process.
4. Compatible with the recent ensembling code.

These are the changes need to be fixed:
1. I hacked around the transformer decoder because I don't know how to fix it. Maybe @msperber can help?
2. Fully supported  the beam search with batch input so the final objective of this pull request can be achieved (allowing to do min risk with hyp generated from the beam search).
3. Support multiple sample for reinforce loss (I don't know the correct calculation atm).

philip30:
Although the beam search code already passed the unittest, I'd appreciate if somebody can validate the newly implemented beam search code. I changed it to be more efficient, although with the addition of required implementation for minrisk and reinforce code to be working, the resulting code can be a bit slower than before.

examples/07_load_finetune.yaml
```diff
@@ -68,3 +68,20 @@ exp2-finetune-model: !LoadSerialized
   - path: train.dev_zero
     val: True
 
+exp3-finetune-minrisk: !LoadSerialized
```
msperber: This should probably go into a new, dedicated minrisk example config.

neubig: Agree

test/config/reinforce.yaml
```diff
@@ -0,0 +1,35 @@
+# A standard training run, should almost never break
```
neubig: This comment seems wrong.

xnmt/loss_calculator.py
```diff
@@ -56,81 +58,126 @@ def __call__(self, translator, dec_state, src, trg):
 
     return dy.esum(losses)
 
+def remove_eos(sequence, eos_sym=Vocab.ES):
```
neubig: This seems a bit strange to have here. Could we make it a function of the sequence itself instead?

xnmt/loss_calculator.py
```diff
@@ -56,81 +58,126 @@ def __call__(self, translator, dec_state, src, trg):
...
+    
+    search_outputs = translator.search_strategy.generate_output(translator, initial_state, forced_trg_ids=trg)
+    for search_output in search_outputs:
+      logprob = search_output.logsoftmaxes
+      sample = search_output.word_ids
+      attentions = search_output.attentions
+
+      logprob = dy.esum(logprob) * self.alpha
+      # Calculate the evaluation score
+      eval_score = [0 for _ in range(batch_size)]
```
neubig: Would `np.zeros()` be better here?

xnmt/search_strategy.py
```diff
@@ -2,125 +2,253 @@
...
       src_length (int): length of src sequence, required for some types of length normalization
       forced_trg_ids (List[int]): list of word ids, if given will force to generate this is the target sequence
     Returns:
-      Tuple[List[int],float]: (id list, score)
+      List[SearchOutput]: List of (word_ids, attentions, score, logsoftmaxes)
     """
     raise NotImplementedError('generate_output must be implemented in SearchStrategy subclasses')
 
-class GreedySearch(SearchStrategy):
+class GreedySearch(Serializable, SearchStrategy):
```
neubig: Do you think we even need greedy search anymore? If it'd be useful that's fine, but we could also remove it in favor of just using beam search with beam size 1.

neubig:
If all of the above are checked, @philip30 feel free to merge at your discretion.

xnmt/search_strategy.py
msperber: I had implemented this originally for unit testing purposes. I guess it would be reasonable to move it into the respective unit testing module instead if it's not needed here?

neubig: I could go either way. In that case let's not change it as part of this PR.

xnmt/loss_calculator.py
philip30: done

xnmt/loss_calculator.py
philip30: done

test/config/reinforce.yaml
philip30: done

examples/07_load_finetune.yaml
philip30: done

philip30:
@msperber as I wrote in the comment before, the previous beam_search copies all the output from the current state to the next state via ```list(state.word_ids)``` and ```list(state.attentions)``` this is an expensive O(N) operation. I simply store the output of current state and create a pointer to the previous state (so we can collect the output from </s> to the first word). We did this by 1 O(N) loop at the end.

I temporarily disable all the unnecessary output collection which can be expensive atm. And I feel that we should have a global flag system that can be modified globally and requested also globally. This allow us to control whether one operation is needed or not (given that a specific object may requires it). The quick hack can be modifying the exp_global directly, but I think we should think more carefully about this operation. (I leave this for future PR).

msperber:
I think using global flags might be problematic because we usually want to activate some of the things only during training and not evaluation, or only for some tasks in a multitask training, etc. It would probably be better to just add a boolean argument to BeamSearch.__init__(), and optionally use parameter sharing or something similar to how vocabs are determined in some places to set that argument automatically if we need the extra convenience (or simply throw an error and remind people that this option needs to be activated manually).

philip30:
#178

Issue #285, opened 2018-03-03
SimpleWordEmbedder can't be referenced in vocab_projector to not having _call_ implemented


neilzxu:
When training using something like 
```
src_embedder: !SimpleWordEmbedder
      _xnmt_id: src_emb
      emb_dim: 512
...
decoder: !MlpSoftmaxDecoder
      layers: 1
      bridge: !CopyBridge {}
      vocab_projector: !Ref { name: src_emb }
```

Gets this error:
```
Traceback (most recent call last):
  File "/home/ziyux/src/xnmt/xnmt/xnmt_run_experiments.py", line 110, in <module>
    sys.exit(main())
  File "/home/ziyux/src/xnmt/xnmt/xnmt_run_experiments.py", line 89, in main
    eval_scores = experiment(save_fct = lambda: yaml_serializer.save_to_file(model_file, experiment,
  File "/home/ziyux/src/xnmt/xnmt/experiment.py", line 35, in __call__
    self.train.run_training(save_fct = save_fct)
  File "/home/ziyux/src/xnmt/xnmt/training_regimen.py", line 103, in run_training
    loss = self.training_step(src, trg)
  File "/home/ziyux/src/xnmt/xnmt/training_task.py", line 244, in training_step
    standard_loss = self.model.calc_loss(src, trg, self.loss_calculator)
  File "/home/ziyux/src/xnmt/xnmt/translator.py", line 133, in calc_loss
    model_loss.add_loss("mle", loss_calculator(self, dec_state, src, trg))
  File "/home/ziyux/src/xnmt/xnmt/loss_calculator.py", line 27, in __call__
    return self.loss_calculator(translator, dec_state, src, trg)
  File "/home/ziyux/src/xnmt/xnmt/loss_calculator.py", line 46, in __call__
    word_loss = translator.decoder.calc_loss(dec_state, ref_word)
  File "/home/ziyux/src/xnmt/xnmt/decoder.py", line 149, in calc_loss
    scores = self.get_scores(mlp_dec_state)
  File "/home/ziyux/src/xnmt/xnmt/decoder.py", line 141, in get_scores
    return self.vocab_projector(h_t)
TypeError: 'SimpleWordEmbedder' object is not callable
```

philip30:
vocab projector is the last layer that project context to vocabulary sized vector, so it is not word embedding. Is this intended?

neilzxu:
[...]

PR #286, opened 2018-03-03
fixed SimpleWordEmbedder can't be passed by reference

neilzxu:
Fix for #285 .


msperber:
``DenseWordEmbedder`` already implements this, with the difference that it also adds a bias for more consistency with the standard vocab projector. Could you check the component sharing example under https://github.com/neulab/xnmt/blob/master/test/config/component_sharing.yaml to see if it fits your needs?

neilzxu:
Ok, great thanks - that makes sense.

PR #288, opened 2018-03-06
Add sequence accuracy as an evaluation metric

mbollmann:
I did not see a way to evaluate simple sequence accuracy, i.e., the percentage of hypothesis sequences that exactly match the references.

This may not be too relevant for machine translation, but can be for other sequence-to-sequence tasks, and seems basic enough that it should exist.

If I missed something and this already exists, please let me know.


neubig:
Thanks! I was actually surprised that we hadn't added this yet, but this will be nice to have.

PR #291, opened 2018-03-07
[WIP] add reranking to inference

cindyxinyiwang:
this commit adds reranking to inference.py, and adds a example config rerank.yaml file


examples/data/head.nbest.en
```diff
@@ -0,0 +1,18 @@
+0 ||| can you do it in one day ?
```
neubig: It'd be nice if we could modify this a little bit to show that these are in fact different hypotheses for the same source sentence.

examples/12_rerank.yaml
```diff
@@ -0,0 +1,75 @@
+# Using a trained model to do reranking score for an nbest list
+
+exp1-model: !Experiment
+  exp_global: !ExpGlobal
+    # The model file contain the whole contents of this experiment in YAML
```
neubig: Delete this comment?

examples/12_rerank.yaml
```diff
@@ -0,0 +1,75 @@
+# Using a trained model to do reranking score for an nbest list
+
```
neubig: Maybe make this a little bit more verbose? What is the location of the existing model, where is the n-best list? What will be the expected output after running this file?

Also, perhaps we should make a distinction in the terminology. Are we doing reranking (taking in an n-best list and picking the best one) or hypothesis scoring (taking in the hypotheses and assigning a score to each)? I think both are useful, so perhaps we can distinguish between the two.

examples/12_rerank.yaml
```diff
@@ -0,0 +1,75 @@
...
+exp1-model: !Experiment
+  exp_global: !ExpGlobal
+    # The model file contain the whole contents of this experiment in YAML
+    # format. Note that {EXP} expressions are left intact when saving.
+    model_file: examples/output/{EXP}.mod
+    log_file: examples/output/{EXP}.log
+    default_layer_dim: 64
+    dropout: 0.5
+    weight_noise: 0.1
```
neubig: Don't you need a `pretrained_model_file` here if we're loading a pretrained model? Perhaps it'd be good to have exp1 train a model, then exp2 actually perform the reranking.

examples/12_rerank.yaml
```diff
@@ -0,0 +1,75 @@
...
+      hyp_file: examples/output/{EXP}.dev_hyp
+  evaluate:
+    - !AccuracyEvalTask
+      eval_metrics: bleu
+      src_file: examples/data/head.ja
+      ref_file: examples/data/head.en
+      hyp_file: examples/output/{EXP}.test_hyp
+
+exp2-rerank: !Experiment
+  # This will load the contents of the above experiments that were saved to the
```
neubig: Delete this comment, as it's already covered in a separate YAML file.

examples/12_rerank.yaml
```diff
@@ -0,0 +1,75 @@
...
+  # YAML file specified after load:
+  # This will carry out the exact same thing, except that {EXP} is resolved to
+  # a different value (making sure we don't overwrite the previous model),
+  # and except for the things explicitly overwritten in the overwrite: section.
+  # It's possible to change any settings as long as these don't change the number
+  # or nature of DyNet parameters allocated for the component.
+  load: examples/output/exp1-model.mod
+  overwrite: # list of [path, value] pairs. Value can be scalar or an arbitrary object
```
neubig: Similarly.

examples/12_score.yaml
```diff
@@ -0,0 +1,69 @@
...
+      src_file: examples/data/head.ja
+      ref_file: examples/data/head.en
+      hyp_file: examples/output/{EXP}.test_hyp
+
+exp2-score: !Experiment
+  load: examples/output/exp1-model.mod
+  overwrite:
+  - path: exp_global.eval_only
+    val: True
+  - path: model.inference.mode
```
neubig: Maybe instead of overwriting each of these individually, you could overwrite the whole of model.inference (might just be simpler)

examples/12_score.yaml
```diff
@@ -0,0 +1,69 @@
...
+    val: score
+  - path: model.inference.ref_file
```
neubig: Maybe `nbest_file` would be a better name for this?

xnmt/inference.py
```diff
@@ -72,12 +72,30 @@ def __call__(self, generator, src_file=None, trg_file=None, candidate_id_file=No
...
+      if args["mode"] == "score":
+        # ref_file should have nbest format; split to hypothesis only and return a list of source indices
+        nbest_lines = io.open(args["ref_file"], "r", encoding="utf-8").readlines()
+        nbest_hyp_name = args["ref_file"] + ".hyp"
+        nbest_hyp_file = io.open(nbest_hyp_name, "w", encoding="utf-8")
+        nbest_src_corpus = []
+        for nbest in nbest_lines:
+          nbest = nbest.split("|||")
+          nbest = [n.strip() for n in nbest]
+          assert len(nbest) > 1, "When performing scoring, ref_file must have nbest format 'index ||| hypthesis'"
```
neubig: "hypothesis"

xnmt/inference.py
```diff
@@ -116,28 +134,33 @@ def __call__(self, generator, src_file=None, trg_file=None, candidate_id_file=No
...
-    with io.open(args["trg_file"], 'wt', encoding='utf-8') as fp:  # Saving the translated output to a trg file
-      src_ret=[]
-      for i, src in enumerate(src_corpus):
-        # This is necessary when the batcher does some sort of pre-processing, e.g.
-        # when the batcher pads to a particular number of dimensions
-        if self.batcher:
-          self.batcher.add_single_batch(src_curr=[src], trg_curr=None, src_ret=src_ret, trg_ret=None)
-          src = src_ret.pop()[0]
+    if args["mode"] != 'score':
+      with io.open(args["trg_file"], 'wt', encoding='utf-8') as fp:  # Saving the translated output to a trg file
```
neubig: "translated output" -> "scored output"

xnmt/inference.py
```diff
@@ -72,12 +72,30 @@ def __call__(self, generator, src_file=None, trg_file=None, candidate_id_file=No
...
     src_corpus = list(generator.src_reader.read_sents(args["src_file"]))
     # Get reference if it exists and is necessary
-    if args["mode"] == "forced" or args["mode"] == "forceddebug":
+    if args["mode"] == "forced" or args["mode"] == "forceddebug" or args["mode"] == "score":
       if args["ref_file"] == None:
         raise RuntimeError("When performing {} decoding, must specify reference file".format(args["mode"]))
-      ref_corpus = list(generator.trg_reader.read_sents(args["ref_file"]))
-      if self.max_len and any(len(s)>self.max_len for s in ref_corpus):
-        logger.warning("Forced decoding with some targets being longer than max_len. Increase max_len to avoid unexpected behavior.")
+      if args["mode"] == "score":
```
neubig: This is indeed a bit hacky. What about (mostly) deleting this if statement, and then stepping through the n-best list one line at a time below?

xnmt/inference.py
```diff
@@ -116,28 +134,33 @@ def __call__(self, generator, src_file=None, trg_file=None, candidate_id_file=No
...
-      for i, src in enumerate(src_corpus):
-        # This is necessary when the batcher does some sort of pre-processing, e.g.
-        # when the batcher pads to a particular number of dimensions
-        if self.batcher:
-          self.batcher.add_single_batch(src_curr=[src], trg_curr=None, src_ret=src_ret, trg_ret=None)
-          src = src_ret.pop()[0]
+    if args["mode"] != 'score':
+      with io.open(args["trg_file"], 'wt', encoding='utf-8') as fp:  # Saving the translated output to a trg file
+        src_ret=[]
+        for i, src in enumerate(src_corpus):
```
neubig: How about we just step over each line in the target file here, we then read in the first column specifying the index to decide which source sentence we want to encode, then use it to score the target sentence. To read in the target sentence, we can use `generator.trg_reader.read_sent(val)`, which reads in a single sentence. Right now in `input.py` there is no function to read a single sentence, but I think this is something that is obviously necessary, so I think it'd be nice to make all of the `read_sents` calls basically call out to `read_sent()`. I think this also might help reduce code duplications across the various subclasses of `InputReader`.

cindyxinyiwang: input_reader calls a iterate_filtered(filename, filter_id) function on the files. I am not sure what is the best way to separate a read_sent function from this. Do you think we should make read_sents(filename) function call iterate_filtered(filename), and make iterate_filtered(filename) call read_sent() for each line in the file?

neubig: Yeah, that sounds right. `read_sent` will take in a plain text string, and convert it into the appropriate `xxxInput` data structure.

examples/12_score.yaml
cindyxinyiwang: I tried overwrite model.inference object with SimpleInference, but it gives me a serialization error: RecursionError: maximum recursion depth exceeded in comparison

msperber: Hey Cindy, if you could post such a failing example into a debugging branch, I could take a look at it, as I'm planning to extend (/fix?) the load/overwrite mechanism soon anyways.

examples/12_score.yaml
neubig: Once this is fixed I'm happy to merge this! I think the failing example of the maximum recursion depth can be fixed in another commit.

examples/12_score.yaml
cindyxinyiwang: Hi, 
Sorry for the late response! I just added the failing example in branch "yaml_load_error". The yaml file is examples/12_score.yaml

examples/12_score.yaml
cindyxinyiwang: Hi,

I just merged my changes with the latest master commits. Should we merge after the loading error is fixed?

PR #294, opened 2018-03-10
Documentation

msperber:
This adds the doc strings of most of the main classes to the API doc page, and also adds or fixes doc strings in a few places. Also some minor cleanup along the way.

The most important places where doc strings are still missing are GeneratorModel, Reportable, loss builder, loss calculator, as I'm not very familiar with those. Maybe @philip30 can take a look?

Also, Experiment, TrainingRegimen, and a few other places are still missing, I'll add those soon.


msperber:
This would be ready to merge unless there are concerns. It's relatively complete and consistent now, although descriptiveness is still lacking in some places.

neubig:
Go ahead, I can't check now but I assume it's fine!

Issue #106, opened 2017-06-27
Feature Request: Ensembling


neubig:
It would be nice to be able to decode with an ensemble of multiple models.

mbollmann:
Is anyone currently working on this?

I'd be very interested in it, and wondered how difficult it would be to implement. Since I'm not that deep into the code base yet, it would be nice to hear some thoughts or considerations from other contributors.

neubig:
I think it's safe to say that nobody is working on this. I'd say in terms of difficulty, I think probably deciding the design is the hardest part, and once the design is fixed it'd not be too difficult. Probably we'll need to think of what the yaml configuration file would look like for an ensembled model. I'll try to think about this a bit and post a draft, and @msperber might have some ideas as well.

msperber:
I think that would be great, and it would also be a good time to do this now because the infrastructure is stable now. We could aim for something like the following:

Training: train several models such as the one in 01_standard.yaml, we will end up with serialized models m1.mod / m1.mod.data, m2.mod / m2.mod.data, etc.

Test time config:
```
experiment: ..
  model: !EnsembleTranslator
    models:
    - !Load
      mod_file: m1.mod
      path: model # this should only load the 'model' part of the serialized experiment; also compatible with multi-tasking by specifying path: train.tasks.0.model instead
    - !Load
      mod_file: m2.mod
      path: model
  evaluate:
    .. # no changes needed here
```

We would need to address 2 limitations in the serialization process (I'll be happy to help with these):
- The only way to load a previously trained model currently is using the load/overwrite mechanism, which only allows loading a single complete experiment, and then overwriting parts of it. Instead, we would like to load parts of several saved experiments (namely, the 'model' parts) and assemble them into one new experiment. The outlined !Load mechanism could do this, and would be a straightforward-to-implement extension of the current load/overwrite mechanism.
- Currently, pretrained DyNet weights are loaded by first initializing the complete hierarchy, registering DyNet parameters where needed, and then finally filling the allocated parameters with pretrained values if available. This requires that the order of initialization is the same when creating the hierarchy initially and when loading a saved model. Extending this requires a bit of thought: either we assume that a whole .data file corresponds to a whole !Load object, and everything is still initialized in the same order below this, the drawback being that we can't apply this to load partial models using !Load that are trained via multi-task training; or, we could assign to each component a named DyNet param subcollection, and passing only the subcollection when initializing those components.

[...]

PR #295, opened 2018-03-13
Ensembling (re #106)

mbollmann:
First draft of a translator object capable of ensemble decoding.

What this tries to do so far:
- Expects individual models to be `DefaultTranslator` instances
- Expects the source and target vocabs of all models to be the same (this is not checked anywhere so far)
- Requires them to be defined at training time (no "load" mechanism yet)

Unfortunately I got stuck while testing this code -- the example file works fine as long as the ensemble only contains a single model, but I couldn't actually test it with more than one:
- If I try to add a duplicate of the `DefaultTranslator` by specifying `- *my_individual_model`, its instantiation fails due to some unresolved reference to the `exp_global` object.
- If I try to add a duplicate of the translator by copying its definition, the model starts training, but crashes at the first dev checkpoint due to a cryptic YAML serialization error I cannot comprehend.

Help appreciated -- or, of course, more general comments if you think the implementation isn't headed in the right direction.


msperber:
Could you try replicating models via the !Ref syntax? That will be a proper reference, while the * syntax creates a copy. If !Ref is not an option, I can take a look tomorrow.

mbollmann:
Okay, so for the !Ref method:
- With the committed code, it complains that `Ref(name=my_individual_model) has not child named src_reader`. This seems to be due to the code in `shared_params()` that I added. (I must admit I'm not really sure how this shared_params thing works or is used.)
- If I remove this code from `shared_params`, YAML serialization again fails at the first dev checkpoint, this time complaining that the list object (holding the individual translators) `has not child named resolved_serialize_params`.

msperber:
It turns out this was a subtle bug in the deserialization process, thanks for spotting it. It should be fixed now.

msperber:
BTW, I tested it with the ```*my_individual_model``` version which is probably the most appropriate.

mbollmann:
I've fixed some bugs and extended the example configuration to show what should work now already.

If you agree with the general approach, the most crucial thing to have would be the model loading feature discussed above, which I hope @msperber could implement/help with?

In any case, I will be able to continue work on this in April, so feel free to continue, modify, or comment away in the meantime.

mbollmann:
@msperber, did you get a chance to look at this yet?

msperber:
Yes, the approach looks good to me! #306 (pending review) should provide all the serialization-related features needed to make this work, so if you want you can already take a look at it.

mbollmann:
My attempt at ensembling three independently trained models is in `xx_ensembling_2.yaml` in the latest commit. Unfortunately, deserialization of the ensemble model fails with
```
ValueError: Could not determine MLP's output size. Please set its vocab_size or vocab member explicitly, or specify the vocabulary of trg_reader ahead of time.
```

`model.decoder.mlp_layer.trg_reader` is a reference to `model.trg_reader`, but apparently it doesn't get resolved as I confirmed that the MLP layer gets `trg_reader == None` during initialization.

@msperber, can you take a look if I'm attempting the model loading correctly? And if yes, maybe run this example config yourself and see where the issue might be?

examples/xx_ensembling_2.yaml
```diff
@@ -0,0 +1,50 @@
...
+      ref_file: examples/data/head.en
+      hyp_file: examples/output/{EXP}.test_hyp
+
+exp2-single: *my_experiment
+exp3-single: *my_experiment
+
+exp4-ensemble: !Experiment
+  model: !EnsembleTranslator
+    src_reader: !Ref {name: my_src_reader}
+    trg_reader: !Ref {name: my_trg_reader}
```
msperber: These references must be by path, i.e. ```!Ref{path: model.models.0.src_reader}``` and ```!Ref{path: model.models.0.trg_reader}```. The reason is that component names are not written out (one reason for not writing them out is to avoid name clashes, as would otherwise occur in this example because saved models 1..3 all have components of the same name).

msperber:
I added one inline comment, but the main problem is that the MLP is not saving its vocab size, and when reloading it into the ensemble the default references that determine the vocab size don't exist anymore because paths have changed.

To fix that, you can add the following:
```self.overwrite_serialize_param("vocab_size", self.output_dim)```
in mlp.py, under line 51 (i.e. under ```self.output_dim = self.choose_vocab_size(vocab_size, vocab, trg_reader)```).

**Comment by msperber** [comment]
there were actually some more issues with Linear and MLP that might cause problems but are addressed in #317

**Comment by mbollmann** [comment]
Thanks @msperber, those comments were really helpful! The example configuration now runs, although with the toy example it's hard to tell if it's doing anything useful.

Therefore, here's my list of things left to do for this PR:

1) Run a benchmark with ensembling to verify it works as intended.
2) Add tests.
3) Add documentation.
4) Implement parameter checking in the EnsembleTranslator, i.e., let it check if the individual models can actually be joined up.

Further features that may be desirable, but maybe not critical:

1) Add different methods of ensembling (on top of the "averaging" method that's currently used).
2) Support ensembling with different checkpoints of the same model. I briefly looked into this, but I think it's not currently possible. The load path for the serialized model's data is set [here](https://github.com/neulab/xnmt/blob/master/xnmt/persistence.py#L827), but I believe customizing this is not enough as the model components from different checkpoints will all have the same ID, and therefore be treated as the same objects -- does that sound about right, @msperber?

Comments?

**Comment by msperber** [comment]
Regarding checkpoint ensembling: you are right the current serialization strategy does not allow this because all components have the same IDs across saved checkpoints. It's probably not worth changing this behavior, but a simple way to overcome this would be doing some simple string replacement: extract all ```xnmt_subcol_name``` values from the YAML file, map them to a new identifier, and replace all strings in the YAML and .data files correspondingly. This could be done by a standalone script, for example.

This is just to say it would be possible with relatively little effort, not that it would be absolutely necessary.

**Comment by mbollmann** [comment]
@msperber, I went with your idea to do checkpoint ensembling and [created a small script for that](https://gist.github.com/mbollmann/0b5fcf712eacc52315ae2517a15eea46).

With this, I ran the IWSLT benchmark, except that I saved 3 checkpoints, ran the script to convert them to standalone models, and evaluated all checkpoints plus an ensemble of them (on tst2013 only):

```
iwslt-evaluate                | BLEU4: 0.271123259111983, 0.616320/0.357324/0.218921/0.137250 (BP = 0.950604, ratio=0.95, hyp_len=32058, ref_len=33682)
iwslt-evaluate-1              | BLEU4: 0.2719343481283306, 0.619245/0.359645/0.221198/0.138321 (BP = 0.946483, ratio=0.95, hyp_len=31926, ref_len=33682)
iwslt-evaluate-2              | BLEU4: 0.2687115560821839, 0.617898/0.357862/0.217959/0.136019 (BP = 0.944354, ratio=0.95, hyp_len=31858, ref_len=33682)
-----------------------------------------------------------------------
iwslt-ensemble                | BLEU4: 0.27751354923194543, 0.614311/0.358778/0.220928/0.139259 (BP = 0.967079, ratio=0.97, hyp_len=32591, ref_len=33682)
```

Looks like it's working as intended!

neubig:
When you think this is ready to merge I'll be happy to do a code review, just @ me!

msperber:
I took a look also, looks great!

PR #296, opened 2018-03-13
Speech feature extraction

msperber:
- speech-feature extraction so we no longer depend on Kaldi or other tools
- small refactoring of preprocessing so that each preprocessing feature is now controlled by its own class
- changed from NPZ to H5 to represent features because it loads much faster (NPZ still supported but not recommended)
- speech examples use the preprocessor to create features based on a small example .wav file.
- added support for custom placeholders so that things like data paths don't have to be repeated throughout the config file

I'm currently testing a TEDLIUM recipe which I'll send in a later PR, but things look almost identical to Kaldi-generated features so far.


neubig:
Basically looks good to me!

neubig:
(Go ahead and merge after fixing conflicts)

Issue #289, opened 2018-03-07
Separate the MlpSoftmaxDecoder's components


mbollmann:
I wondered how difficult it would be to separate the LSTM and MLP components of `MlpSoftmaxDecoder`.  My motivation for this is setting up multi-task experiments where the tasks share the decoder LSTM, but not the MLP.

At first I tried to just implement a quick hack for myself, but it turned out to be trickier than I thought.  Therefore, I wanted to ask for opinions and if there's interest to do this "the proper way". Particularly:

1. Could the LSTM and MLP parts of `MlpSoftmaxDecoder` just be encapsulated in their own serializable objects, which then become attributes of the decoder?
2. Couldn't the LSTM part of the decoder re-use `UniLSTMSeqTransducer`? It seems to me that the underlying model should be equivalent to what the decoder uses by default, but I don't know about all the implications for the implementation or serialization.
3. Do you think this would be a difficult change to make, or do you maybe have suggestions for a temporary workaround I could use?

Thanks for any input!

msperber:
Yes, this would definitely be a good idea and in the spirit of XNMT. It shouldn't be hard to do, either. You would basically replace ```rnn_spec```, ```residual_to_output```,``` layers```, etc. by a single rnn argument, which could be set to any RNN builder, and the same for the MLP.

Note that it wouldn't be possible to pass ```dy.CompactVanillaLSTMBuilder``` as the RNN, because it is not serializable, and ```UniLSTMSeqTransducer``` is not usable as is because it can currently only transform a complete sequence, not step by step. The cleanest solution would be to implement ```initial_state()``` and ```add_input()``` for ```UniLSTMSeqTransducer```, similar as it is done in residual.py. A slightly faster way would be to make a shallow wrapper for ```dy.CompactVanillaLSTMBuilder``` that basically passes on all method calls to the dynet builder, and only adds Serializable as a base class.

One slight complication might be setting defaults for the rnn and mlp, because some arguments depend on whether or not input feeding is used, so the normal get_shared_params() mechanism probably won't suffice. I think it'd be worth to ignore that for now (meaning we would require to specify rnn and mlp explicitly in the config file as a first approximation), and think of a solution later.

Hope that helps! Feel free to post a [WIP] PR if you get stuck so I can take a look.

mbollmann:
[...]

PR #298, opened 2018-03-14
[WIP] Separate the MlpSoftmaxDecoder's components (re #289)

mbollmann:
This is WIP for separating the decoder components as discussed in #289.

Currently, this code:
- Implements multiple layers and state functionality for `UniLSTMSeqTransducer`
- Refactors the `MlpSoftmaxDecoder` to take an `UniLSTMSeqTransducer` directly

What's missing:
- Get default arguments to work (see below) so that the implementation can be continued
- Refactor residual RNN to work as an RNN layer for the decoder
- Refactor the MLP components into a separate object as well
- Make sure all examples, tests etc. are adapted to the new architecture

As discussed in #289, default arguments don't work right now due to input feeding. That means a lot of tests fail simply because of that. I would strongly prefer finding a solution to this before continuing this WIP.

I was thinking to explicitly make `UniLSTMSeqTransducer` aware of input feeding during `__init__`, so that it can modify its default dims accordingly. However, the main roadblock I encountered is that I don't see a way to detect (within `UniLSTMSeqTransducer`) whether it's child of a decoder or not.

Anyway, comments/suggestions welcome.


msperber:
That's a good idea to have ```UniLSTMSeqTransducer``` try to detect if it's a child of a decoder. It should be possible similar to how the vocab size is determined, i.e. by taking yaml_path in the constructor which will be set automatically when the YAML file is loaded: https://github.com/neulab/xnmt/blob/master/xnmt/embedder.py#L57

xnmt/decoder.py
```diff
@@ -47,49 +50,52 @@ class MlpSoftmaxDecoder(RnnDecoder, Serializable):
...
-    if not isinstance(param_init_lstm, GlorotInitializer): raise NotImplementedError("For the decoder LSTM, only Glorot initialization is currently supported")
-    if getattr(param_init_lstm,"gain",1.0) != 1.0:
-      for l in range(layers):
-        for i in [0,1]:
-          self.fwd_lstm.param_collection().parameters_list()[3*l+i].scale(param_init_lstm.gain)
-      
+    self.rnn_layer = rnn_layer
+
+    #=== TODO: what does this do and why is it needed? how to preserve it with
+    #===       the refactoring?
```
msperber: You can safely delete this. The DyNet RNN builders are hard to configure regarding parameter initialization, so this was a kind of workaround, but with the XNMT builders parameter initialization is fully supported and already implemented.

mbollmann:
Thanks @msperber, utilizing `yaml_path` works like a charm!

mbollmann:
This should be functional now!

Some notes:
1. The decoder instantiation in the tests has become quite verbose, as the `yaml_path` trick doesn't work there automatically to make the sub-components aware they're in a decoder. Currently, I've solved this by adding a `yaml_path` everywhere. This also means I had to explicitly instantiate the decoder components everywhere just for that. If you can think of a more elegant solution here, please go ahead.
2. The `ResidualLSTMSeqTransducer` is not working as a decoder component at the moment. However, I also couldn't make it work on the master branch. The decoder expects some functionality (like `.set_s()`) that just isn't there on the `PseudoState` object. Since I'm not familiar enough with the residual LSTM implementation, I decided to leave this be for now.

From March 18 to 31, I'll be on vacation and unable to do anything with the code, but feel free to continue with the code or leave comments for me to work on when I return -- whatever you prefer.

xnmt/lstm.py
```diff
@@ -9,28 +9,81 @@
...
     param_init (ParamInitializer): how to initialize weight matrices; if None, use ``exp_global.param_init``
     bias_init (ParamInitializer): how to initialize bias vectors; if None, use ``exp_global.bias_init``
   """
-  def __init__(self, exp_global=Ref(Path("exp_global")), input_dim=None, hidden_dim=None,
-               dropout = None, weightnoise_std=None, param_init=None, bias_init=None):
+  yaml_tag = '!UniLSTMSeqTransducer'
+
+  def __init__(self, exp_global=Ref(Path("exp_global")), layers=1, input_dim=None, hidden_dim=None,
+               dropout=None, weightnoise_std=None, param_init=None, bias_init=None,
+               yaml_path=None, decoder_input_dim=None, decoder_input_feeding=True):
```
msperber: The new arguments should be documented.

xnmt/mlp.py
```diff
@@ -1,10 +1,75 @@
...
+    output_dim (int): output dimension; if None, use ``exp_global.default_layer_dim``
+    param_init_hidden (ParamInitializer): how to initialize hidden weight matrices; if None, use ``exp_global.param_init``
+    bias_init_hidden (ParamInitializer): how to initialize hidden bias vectors; if None, use ``exp_global.bias_init``
+    param_init_output (ParamInitializer): how to initialize output weight matrices; if None, use ``exp_global.param_init``
+    bias_init_output (ParamInitializer): how to initialize output bias vectors; if None, use ``exp_global.bias_init``
+    output_projector: TODO
+    vocab_size (int): vocab size or None; only relevant if MLP is used as a decoder component
+    vocab (Vocab): vocab or None; only relevant if MLP is used as a decoder component
+    trg_reader (InputReader): Model's trg_reader, if exists and unambiguous; only relevant if MLP is used as a decoder component
+  """
```
msperber: It should be mentioned that vocab size etc. will overwrite output_dim if given.

neubig:
@msperber If you think it's OK, go ahead and merge. I trust your judgement and think we only need one reviewer per PR.

xnmt/lstm.py
mbollmann: `yaml_path` as well, even though it's kind of for internal use? Because the `decoder_...` arguments only get interpreted when `"decoder" in yaml_path` ...

mbollmann:
I think a benchmark is a good idea, but I don't have any prepared benchmarks or anything that I could easily run at the moment. Any immediate recommendations come to mind?

xnmt/lstm.py
msperber: Hm, maybe you could add ```yaml_path``` as an argument but leave the description empty? And then somehow the information should be conveyed that ```decoder_...``` is only interpreted when ```yaml_path``` is given.

msperber:
The easiest to use would probably be the Stanford benchmark because it's already preprocessed: https://nlp.stanford.edu/projects/nmt/
I often use TED because it's smaller and fast to train a model, but you'd have to do the preprocessing yourself: https://wit3.fbk.eu/mt.php?release=2017-01-trnted

If you can somehow guarantee that both models are initialized with the exact same random parameters, it would also be enough to run just one training epoch plus evaluation and make sure scores are exactly the same.

mbollmann:
> The only minor difference to the master branch is that the decoder LSTM is initialized slightly differently now: with UniLSTMTransducer, the Glorot initializer treats the 4 weight matrices as separate matrices now instead of one bigger matrix as was the case with the dynet builder.

I think this might actually be a good thing, as it makes the initialization consistent with the encoder LSTM. However, I'm not sure what the easiest way would be to get the exact same random initialization for both the old and new architecture, as simply fixing the random seed won't suffice then.

Either way, I don't think I'll get around to this (and the benchmarks in general) before my vacation. If you want to take this up, please go ahead, otherwise I can continue that in April.

msperber:
Yes, it's definitely a good thing! Getting the parameters to be identical would probably some custom parameter saving / loading code, so I'd probably start with the complete benchmark as it involves less human labor (although more computer labor) :) In any case, I might also give it a shot depending on when I get around to the recipes.

msperber:
To follow up on this, I haven't had a chance to actually run it, but I would suggest the new stanford-iwslt recipe which uses a small data set and requires minimal effort to run (basically run a download script and then start training).

mbollmann:
Oh, that's cool, I actually downloaded that very dataset an hour ago and am looking for a machine to run it on now. ;)

msperber:
Sorry also for the merge conflicts, let me know if you need help resolving them.

mbollmann:
Results for a single run of the IWSLT recipe on master:

```
Experiment                    | Final Scores
-----------------------------------------------------------------------
iwslt-experiment              | BLEU4: 0.2475964750860351, 0.586186/0.332474/0.197394/0.120693 (BP = 0.948510, ratio=0.95, hyp_len=32575, ref_len=34297)
                              | BLEU4: 0.27212110133216566, 0.623154/0.360932/0.221823/0.140138 (BP = 0.941058, ratio=0.94, hyp_len=31753, ref_len=33682)
```

And on this feature branch: 
```
Experiment                    | Final Scores
-----------------------------------------------------------------------
iwslt-experiment              | BLEU4: 0.24962836908499383, 0.584297/0.329941/0.194742/0.118162 (BP = 0.967257, ratio=0.97, hyp_len=33192, ref_len=34297)
                              | BLEU4: 0.2732730662351511, 0.618037/0.357961/0.219796/0.138445 (BP = 0.954025, ratio=0.96, hyp_len=32168, ref_len=33682)
```

msperber:
Great, thanks! Would you mind adding these numbers (the new ones) to the recipe's README file? Otherwise, I think we can merge once the remaining failing test is resolved.

neubig:
Looks good! Could we set the the following:

    rnn_layer: !UniLSTMSeqTransducer
      layers: 1

to be the default and remove it from the config files, unless it would be useful to have for illustrative purposes? Maybe similarly for the MLP layer?

Also, this could maybe be for another commit, but we could potentially generalize `UniLSTMSeqTransducer` to `UnidirectionalSeqTransducer` that defines the interface for sequence transducers that can add elements one-by-one left to right.

mbollmann:
@neubig, that is the default already, so it's actually safe to remove. But then again, there's a lot of redundancy in the example configs anyway with the explicit dimensionality definitions (even though they usually match `default_layer_dim`).

I noticed that there are a few new examples that I hadn't adapted yet, so I'm also taking care of those now.

mbollmann:
Unless I missed some config file to adapt, or we want to remove more of the `rnn_layer` definitions from them, this should be ready for merging from my side.

xnmt/decoder.py
```diff
@@ -46,34 +37,21 @@ def __init__(self, rnn_state=None, context=None):
...
-    dropout (float): dropout probability for LSTM
-    rnn_spec (str): 'lstm' or 'residuallstm'
-    residual_to_output (bool): option passed on if rnn_spec == 'residuallstm'
     input_feeding (bool): whether to activate input feeding
-    param_init_lstm (ParamInitializer): how to initialize LSTM weight matrices (currently, only :class:`xnmt.param_init.GlorotInitializer` is supported)
-    param_init_context (ParamInitializer): how to initialize context weight matrices
-    bias_init_context (ParamInitializer): how to initialize context bias vectors
-    param_init_output (ParamInitializer): how to initialize output weight matrices
-    bias_init_output (ParamInitializer): how to initialize output bias vectors
+    rnn_layer (SeqTransducer): recurrent layer of the decoder; defaults to UniLSTMSeqTransducer
```
msperber: SeqTransducer is not the correct type here, as SeqTransducers can't consume inputs one by one. For lack of an appropriate abstract class, the requested type should probably be UniLSTMSeqTransducer.

msperber:
I took another look and had one mini comment but other than that we can merge this.

xnmt/decoder.py
mbollmann: Yeah, that's where @neubig's suggestion would probably come in handy. Also, it's not like you should really use the BiLSTMSeqTransducer either, so...

msperber:
Cool, will merge!

PR #301, opened 2018-03-17
Basic config type checking

msperber:
This is a suggestion to perform basic type checking of values specified in YAML configs. This will catch some common errors such as ```overwrite: Fals``` being interpreted as equal to ```True``` (the spelling mistake makes this a string rather than bool).

It uses Python type annotations, so I moved some of the type documentation from the docstring to the method definition. Either is fine according to the Google documentation guidelines, although it might be a bit confusing for people not familiar with this, so let me know if you have reservations.

As a limitation, this currently only checks "normal" types (like ```bool``` or ```Experiment```), not things from the typing module (like ```List[int]``` or ```Optional[Experiment]```), which means it can currently only catch some of the simpler cases but could be extended later if necessary.

### DIFF
--- a/xnmt/attender.py
+++ b/xnmt/attender.py
@@ -112,7 +112,7 @@ class DotAttender(Attender, Serializable):
 
   yaml_tag = '!DotAttender'
 
-  def __init__(self, scale=True):
+  def __init__(self, scale:bool=True):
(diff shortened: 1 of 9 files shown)


neubig:
This is a good idea!

PR #302, opened 2018-03-18
Yaml load error

cindyxinyiwang:
This branch contains a failed example of model loading. Simply run with command `python -m xnmt.xnmt_run_experiments examples/12_score.yaml `. The error message is "RecursionError: maximum recursion depth exceeded in comparison"


neubig:
Probably fixed by #306

PR #306, opened 2018-03-28
[WIP] Improved Serializable Interface

msperber:
This improves the Serializable interface as follows:
- consistent results between creating components from YAML or creating them programmatically. When created programmatically, references are attempted to replaced by default values, or an Exception is raised. bare objects are automatically instantiated.
- DyNet parameters are now managed globally. Each class is assigned their own unique subcollection, which will make it possible to assemble models from different saved models.
- Usage of global dropout and default_layer_dim, etc. is now made explicit in the class signature.
- remove need to pass exp_global into every constructor.

Once finished, this will allow ensembling, as well as "model assembly" by combining components that were independently trained into a new model.

This requires some slightly new conventions, which I've explained under "Writing XNMT Classes" in the documentation. In short:
- ```__init__``` of Serializable classes must be decorated with ```@serializable_init```
- Subobjects that are Serializable must be created using the self.reuse_or_register() helper.
- exp_global can no longer be passed, instead its properties should be referenced directly.

The load/overwrite mechanism is replaced by !LoadSerialized object, which can be placed anywhere in the hierarchy and load subcomponents.

- [X] Named subcollections for each component
- [X] Consistent interface, example for using XNMT programmatically (without YAML config file).
- [ ] Fully implement !LoadSerialized


msperber:
This should be ready now. There's an example under test/config/assemble.yaml that shows how to combine several models into a new one. I haven't made an "official" example because I'd still consider it experimental code; the ensembling feature will be a good practice test.

neubig:
I haven't had time to look at it closely, but I think it's basically good so you can merge if you think it's working. Here are some naming suggestions, in the general spirit of wanting the code to be as readable as possible without knowing xnmt's coding conventions (something we should strive for):

* `my_subcollection` -> `my_params`
* `reuse_or_register` -> `add_component` or `add_serializable_component`

PR #307, opened 2018-03-29
Preproc: new normalizer that removes punctuation

msperber:
(+ small fix in preproc example config)

### DIFF
--- a/xnmt/preproc.py
+++ b/xnmt/preproc.py
@@ -58,7 +58,7 @@ class NormalizerRemovePunct(Normalizer):
     else:
       words = []
       for w in sent.split():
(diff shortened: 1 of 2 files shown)


xnmt/preproc.py
```diff
@@ -34,6 +35,8 @@ def from_spec(spec):
       for my_spec in spec:
         if my_spec["type"] == "lower":
```
neubig: Just a comment: this specification of arguments via strings doesn't follow the current xnmt paradigm of specifying everything through classes, and could be fixed. In fact the whole preproc section might be worth refactoring. That can wait for another PR though.

PR #308, opened 2018-03-29
Implement word_truncate for H5 / Npz input readers

msperber:
This truncates vector input sequences to some given maximum length.


xnmt/input_reader.py
```diff
@@ -198,12 +198,13 @@ class H5Reader(InputReader, Serializable):
   """
   yaml_tag = u"!H5Reader"
 
-  def __init__(self, transpose=False, feat_from=None, feat_to=None, feat_skip=None, word_skip=None):
+  def __init__(self, transpose=False, feat_from=None, feat_to=None, feat_skip=None, word_skip=None, word_truncate=None):
```
neubig: Because I'm not familiar with this class I'm not sure what this PR is doing. Could you document the `__init__` function?

msperber:
Sure, I added some documentation.

xnmt/input_reader.py
```diff
@@ -190,11 +190,12 @@ class H5Reader(InputReader, Serializable):
...
-    transpose (bool):
-    feat_from (int):
-    feat_to (int):
-    feat_skip (int):
-    word_skip (int):
+    transpose (bool): whether inputs are transposed or not.
+    feat_from (int): use feature dimensions in a range, starting at this index (inclusive)
+    feat_to (int): use feature dimensions in a range, ending at this index (exclusive)
+    feat_skip (int): stride over features
+    word_skip (int): stride over words
```
neubig: Are we assuming that each column in the feature vector corresponds to a "word" with this notation? Maybe "time", "timestep", or something else would be better?

msperber: Sure, thanks for the hint!

PR #310, opened 2018-04-04
update score example and also add as unit test

msperber:
This converts the config file of #302 to the most recent format and confirms that it's working by adding a unit test. Probably #306 fixed the issue of #302.


neubig:
Thanks!

PR #312, opened 2018-04-04
enable zero-evaluation (before training starts)

msperber:
This is useful e.g. when finetuning a pre-trained model.


msperber:
Let me withdraw this actually, sorry for the premature PR. It will make a lot more sense to make a dev_zero version instead so that a finetuning experiment will write out the pretrained model as best model in case the finetuning itself actually didn't improve things.

PR #313, opened 2018-04-04
Zero dev checkpoint

msperber:
Optionally add a dev checkpoint before first training step. This is useful for monitoring finetuning experiments. Replaces #312


neubig:
LGTM

PR #316, opened 2018-04-06
Refactor scoring

msperber:
This is partly a code cleanup, partly making evaluation more flexible:
- xnmt_evaluate via command line mode now takes either a string like ```bleu,gleu```as metrics, or a fully configurable list of python object such as ```WEREvaluator(case_sensitive=True),BLEUEvaluator()]```. Previously, the latter was not possible, so evaluators could only be used with default settings. 
- the same holds for the metrics of the ```AccuracyEvalTask```.

I do think the ```bleu,gleu``` string shortcut is a good idea for the command line because we don't want to specify python objects if we don't have to, and the required extra code is quite simple. The main question is whether this is true for YAML configs also: do we want to support ```eval_metrics:wer,bleu``` or only ```eval_metrics: [!WEREvaluator {}, !BLEUEvaluator {}]```? And if we support the former, do we want to use the shortcuts or the more consistent YAML style in the example configs?

### DIFF
--- a/xnmt/evaluator.py
+++ b/xnmt/evaluator.py
@@ -7,6 +7,7 @@ import subprocess
 import numpy as np
 
 from xnmt.serialize.serializable import Serializable
+from xnmt.serialize.serializer import serializable_init
 
 class EvalScore(object):
   def higher_is_better(self):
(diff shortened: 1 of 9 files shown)


neubig:
I think supporting the string shortcuts is a good idea, and in the yaml config for simplicity maybe use the string shortcuts.

msperber:
Ok, in that case this would be ready to merge from my side.

xnmt/evaluator.py
```diff
@@ -536,52 +541,33 @@ def evaluate(self, ref, hyp):
...
+  def evaluate(self, ref, hyp, desc=None):
     true_positive = 0
     for hyp_i, ref_i in zip(hyp, ref):
       if any(ref_i == idx for idx, _ in hyp_i[:self.nbest]):
         true_positive += 1
     score = true_positive / float(len(ref))
-    return RecallScore(score, len(hyp), len(ref), nbest=self.nbest, desc=self.desc)
+    return RecallScore(score, len(hyp), len(ref), nbest=self.nbest, desc=desc)
 
-class MeanAvgPrecisionEvaluator(object):
```
neubig: Why was this deleted? If it's incompatible, let's just comment it out for now and restore it later. It's necessary for evaluating retrieval models.

PR #317, opened 2018-04-09
Better Persistence

msperber:
- fix an issue that caused Linear and MLP to be not saved properly
- make YAML (de-)serialization a bit less confusing and reduce the chance of introducing bugs:
  - imports.py is moved to package init and no longer needs to be imported manually
  - other modules are merged into a new "persistence.py" module so stuff doesn't have to be imported from multiple modules (division was not super clean in the first place).
  - OptionParser is renamed -> YamlPreloader
  - YamlSerializer no longer needs to be instantiated by hand
  - instead, the main functions for loading / saving are now module-level: initialize_if_needed(), initialize_object(), save_to_file()
  - Serializable.shared_params() can now return sets of strings instead of sets of Path objects
  - Serializable.overwrite_serialize_param is renamed -> save_processed_arg()
  - some more bug checks, informative error messages, doc in some places


philip30:
This is a very good refactoring. I have 1 suggestion. How about if we also move the logger to the package init.

So instead of calling 2 lines at every file like this:

```
import logging
 logger = logging.getLogger('xnmt')
```

We can move those lines to ```__init__.py``` and use this line to import at every file

```
from xnmt import logger
```

We can, of course, create several logger objects if needed (e.g. for yaml logger)

msperber:
Thanks, that sounds like a good idea!

neubig:
I don't fully understand the serialization interface at the moment, but I'm also probably not going to have time to check it carefully soon, so I'm happy to have you go ahead and merge. I browse quickly and think this looks fine (and also like @philip30 's suggestion).

PR #323, opened 2018-04-12
Length norm improvements

msperber:
- fix length normalization not working in case of ```apply_during_search=False```
- speed up ```PolynomialNormalization``` by precomputing powers of sequence lengths that are otherwise recomputed a large number of times. Overall speedup of beam search is about 10%.


neubig:
Thanks!

PR #324, opened 2018-04-12
dev_every: remove special handling of new epoch events

msperber:
Currently, checkpoints are made whenever an epoch is finished, or more often in case ```dev_every``` is set.  It is not possible to have checkpoints less often than at every epoch though, even if setting ```dev_every``` to a value greater than the number of training sentences. This commit changes this behavior: finishing an epoch has no special significance if ```dev_every``` is set, so that we can set it so that checkpoints are less often than once every epoch (useful e.g. in connection with ```sample_train_sents```).


neubig:
Good idea!

Issue #322, opened 2018-04-11
Problems with the provided recipes


pmichel31415:
I'm trying to run the iwslt en-vi recipe in https://github.com/neulab/xnmt/tree/master/recipes/stanford-iwslt

But I get several errors:

1. I had to replace the command 
```bash
python -m xnmt.xnmt_run_experiments --dynet-gpu ./config.en-vi.yaml
```
with 
```bash
python -m xnmt.xnmt_run_experiments --dynet-gpu --settings=config.en-vi.yaml
```
because it would fall back to `settings.standard`
2. Even with this I get the following error message
```bash
Traceback (most recent call last):
  File "/home/pmichel1/anaconda3/lib/python3.6/runpy.py", line 183, in _run_module_as_main
    mod_name, mod_spec, code = _get_module_details(mod_name, _Error)
  File "/home/pmichel1/anaconda3/lib/python3.6/runpy.py", line 109, in _get_module_details
    __import__(pkg_name)
  File "/home/pmichel1/.local/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/__init__.py", line 25, in <module>
    import xnmt.attender
  File "/home/pmichel1/.local/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/attender.py", line 6, in <module>
    from xnmt.param_init import GlorotInitializer, ZeroInitializer
  File "/home/pmichel1/.local/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/param_init.py", line 6, in <module>
    from xnmt.persistence import serializable_init, Serializable
  File "/home/pmichel1/.local/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/persistence.py", line 18, in <module>
    from xnmt.tee import get_git_revision
  File "/home/pmichel1/.local/lib/python3.6/site-packages/xnmt-0.0.1-py3.6.egg/xnmt/tee.py", line 39, in <module>
    logger.setLevel(min(logging._checkLevel(settings.LOG_LEVEL_CONSOLE), logging._checkLevel(settings.LOG_LEVEL_FILE)))
  File "/home/pmichel1/.local/lib/python3.6/site-packages/simple_settings/core.py", line 84, in __getattr__
    self.setup()
  File "/home/pmichel1/.local/lib/python3.6/site-packages/simple_settings/core.py", line 65, in setup
    self._load_settings_pipeline()
[...]

PR #326, opened 2018-04-13
Remove simple settings

msperber:
simple_settings was problematic in some situations (see #322 e.g.), so this is getting rid of this dependency and implements a simple custom version that doesn't attempt loading external modules and falls back to standard settings if no settings are specified.


neubig:
Thanks! This was a source of bugs for me frequently as well, so it'll be nice to have the custom version.

Issue #329, opened 2018-04-16
Maybe librosa is unnecessary?


neubig:
librosa seems to be a bit of a specialized library and it's not clear to me that we should have this as a requirement for xnmt itself. Could we make librosa only be imported from utility (e.g. preprocessing) scripts, and remove it from requirements.txt for xnmt itself? @msperber

PR #331, opened 2018-04-16
Removed unnecessary requirements

neubig:
Fixes https://github.com/neulab/xnmt/issues/329

Looks like all tests and training pass without librosa and exact h5 version, so here is a cleanup. Also small addition to gitignore for the test directory.


msperber:
LGTM! Should we maybe put librosa into a new ```requirements-preproc.txt``` or ```requirements-extra.txt```?

neubig:
Thanks! Added.

PR #332, opened 2018-04-16
fix case of exp_dir==WD

### LINKED ISSUE #330: Config file paths need to add `./`
See comment here:
https://github.com/neulab/xnmt/issues/322#issuecomment-380854390

I got stuck on this as well and wasn't able to figure out why for a while, so this should definitely be fixed.

msperber:
fixes #330


neubig:
Thanks!

PR #335, opened 2018-04-17
Lexicon Bias method

philip30:
This is the implementation of the lexicon-bias model.


philip30:
The linear method is implemented as intended but didn't work really well.

xnmt/decoder.py
```diff
@@ -152,3 +161,120 @@ def calc_loss(self, mlp_dec_state, ref_action):
...
+    src_unk_id = self.src_vocab.convert(self.src_vocab.unk_token)
+    trg_unk_id = self.trg_vocab.convert(self.trg_vocab.unk_token)
+    lexicon[self.src_vocab.SS] = {self.trg_vocab.SS: 1.0}
+    lexicon[self.src_vocab.ES] = {self.trg_vocab.ES: 1.0}
+    # TODO(philip30): Note sure if this is intended
+    lexicon[src_unk_id] = {trg_unk_id: 1.0}
+    return lexicon
+
+  @handle_xnmt_event
+  def on_new_epoch(self, training_task, *args, **kwargs):
```
neubig: Why is this in `on_new_epoch` and not in the constructor?

xnmt/decoder.py
```diff
@@ -130,6 +136,9 @@ def get_scores(self, mlp_dec_state):
     """
     return self.mlp_layer(dy.concatenate([mlp_dec_state.rnn_state.output(), mlp_dec_state.context]))
 
+  def get_scores_logsoftmax(self, mlp_dec_state):
```
neubig: What about just `get_scores`? I think it's fine to do `log_softmax` in the parent function if necessary. We might not need to take the log softmax.

xnmt/decoder.py
philip30: I was afraid that the vocabulary has not been populated in case of vocab is constructed when reading the corpus. In this case, the length of the vocabulary will be zero. on_new_epoch will ensure that the vocabulary has been populated so we get the correct length

xnmt/decoder.py
philip30: In the case where we want to modify the probability of the vocabulary directly in the decoder (e.g. linear interpolation), the softmax must be computed in the decoder, not in the parent function.

xnmt/decoder.py
neubig: Ahh, fair enough.

xnmt/decoder.py
neubig: OK, that sounds reasonable.

neubig:
Thanks!

PR #337, opened 2018-04-18
Refactor loss tracker

### LINKED ISSUE #328: Loss tracker has problems with multi task training
There are at least 2 loss tracker issues when using multi task training:
- speed measurements are wrong
- sample_train_sents causes dev checkpoints to be made after every training minibatch

msperber:
This fixes #328 and refactors loss trackers as follows:
- separate class for train and dev loss trackers
- the train loss tracker now belongs to the training regimen, not training task, so that speed measurements can be carried out consistently even for multi task training regimens
- speed measurements are made explicit, by using with statements
- the training state is now completely controlled by training_task.training_state. loss trackers only keep what's necessary for outputting stuff at the proper time, but nothing on which training tasks depend.
- dev logging is improved and fixed for some multitask cases
- save model prior to training and explicitly check for run_for_epochs==0 to avoid problems in some edge cases
- make sure all multi task regimens are captured by unit tests


neubig:
Thanks! I just was about to test the multitasking stuff, so I'll do a stress test.

PR #341, opened 2018-04-18
Doc examples: fix pathing issues

pmichel31415:
Sorry for the abundance of PR.

My previous PR failed on readthedocs because of relative paths, this should fix it (we use the same trick in dynet)


neubig:
Be happy to merge once conflicts are resolved.

pmichel31415:
I'll open another PR instead

Issue #321, opened 2018-04-11
Sentencepiece integration


pmichel31415:
Not sure if that'll be helpful but FWIW sentencepiece now has official python bindings (https://github.com/google/sentencepiece/blob/master/python/README.md), which might be more convenient for several reasons:

- can be installed with pip (no separate installation, just add to `requirements.txt`
- easier on-the-fly tokenization/detokenization

neubig:
Awesome! We should definitely use this, assuming it supports training.

pmichel31415:
It does! I've been using it for another project. Here's what the API looks like:

Training:
```python
import sentencepiece as spm
spm.SentencePieceTrainer.Train('--input=/path/to/corpus/file --model_prefix=my.model.name --model_type=bpe --vocab_size=8000')
```
Tokenizing
```python
import sentencepiece as spm
sp = spm.SentencePieceProcessor()
sp.Load('my.model.name.model')
sp.EncodeAsPieces('a sentence') # ['a', '_sent', 'ence']
```

pmichel31415:
OK this is basically done now

PR #344, opened 2018-04-19
Use sentencepiece python API

pmichel31415:
Addresses #321

### DIFF
--- a/xnmt/preproc.py
+++ b/xnmt/preproc.py
@@ -171,17 +171,24 @@ class ExternalTokenizer(Tokenizer):
 
 class SentencepieceTokenizer(Tokenizer):
   """
-  A wrapper around an independent installation of the sentencepiece tokenizer
-  with passable parameters.
+  Sentencepiece tokenizer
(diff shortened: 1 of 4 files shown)


xnmt/preproc.py
```diff
@@ -200,21 +202,23 @@ def __init__(self, path, train_files, vocab_size, overwrite=False, model_prefix=
...
-          ]
-      subprocess.call(sentpiece_train_command)
-
-    sentpiece_encode_exec_loc = os.path.join(self.sentpiece_path, 'spm_encode')
-    sentpiece_encode_command = [sentpiece_encode_exec_loc
-        , '--model=' + self.model_prefix + '.model'
-        , '--output_format=' + self.output_format
-        ] + self.encode_extra_options
-    self.tokenizer_command = sentpiece_encode_command
+      sentpiece_train_args = ['--input=' + ','.join(train_files),
```
neubig: Could you also add `--hard_vocab_limit=false`. This will keep it from crashing on small corpora.

PR #353, opened 2018-04-20
Reuse sentencepiece vocab file

pmichel31415:
This adds an option to vocab to make it possible to read the vocabulary from the `.vocab` file generated by sentencepiece. This saves a lot of preprocessing.


neubig:
Tests seem to be failing.

Issue #293, opened 2018-03-09
Schedule metric usage unclear


msperber:
Currently, the evaluator that appears first among all dev evaluators is used to determine the schedule (early stopping, patience), which is an important feature but it's not documented anywhere. It should either be documented or be configurable in the config file. In #260 there was a plan to make this even more flexible using something like: 
```
  train: !SimpleTrainingRegimen
    ..
    dev_tasks:
      - !AccuracyEvalTask
        model: *mt_model
        eval_metrics: bleu
        src_file: mt-dev.src
        trg_file: mt-dev.trg
      - !AccuracyEvalTask
        model: *pos_model
        eval_metrics: wer
        src_file: pos-dev.src
        trg_file: pos-dev.trg
    eval_combinator: "x[0]-x[1]"
```

PR #354, opened 2018-04-20
[WIP] Added ability to combine together dev scores

neubig:
This PR introduces the ability to combine together multiple development scores to determine the final score that should be used to do learning rate decay, etc. It also improves the documentation about the default of using the first development task.

The reason why this is "[WIP]" is because using `eval()` is a security hole: https://stackoverflow.com/questions/661084/security-of-pythons-eval-on-untrusted-strings

It'd be nice to be able to come up with a safe version.

Fixes https://github.com/neulab/xnmt/issues/293

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,4 +5,5 @@ dynet
 matplotlib    
 cython   
 h5py
+asteval
 sentencepiece>=0.0.6
--- a/test/config/dev_combinator.yaml
+++ b/test/config/dev_combinator.yaml
@@ -1,0 +1,35 @@
+# A standard training run, should almost never break
+dev_combinator: !Experiment
+  exp_global: !ExpGlobal
+    model_file: examples/output/{EXP}.mod
+    log_file: examples/output/{EXP}.log
+    default_layer_dim: 64
+  model: !DefaultTranslator
+    src_reader: !PlainTextReader
+      vocab: !Vocab {vocab_file: examples/data/head.ja.vocab}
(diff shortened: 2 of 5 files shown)


msperber:
For the record, YAML deserialization is actually also not safe because it allows creating arbitrary Python objects. However, it does have a load_safe method which is more restrictive and could be tried if we needed to.

Instead of eval, I guess we could try a serializable WeightedSum class that has configurable weights (including negative)? It would be much less flexible though, obviously.

And one more remark, I've also used ```eval()``` here, although it could probably be replaced by ```yaml.load()```: https://github.com/neulab/xnmt/blob/master/xnmt/xnmt_evaluate.py#L69

msperber:
YAML also has a ```!!python/object:``` syntax for creating arbitrary objects, so I think you could do something like ```!!python/object:subprocess.Popen{ ... }``` to trick it into executing arbitrary commands. The ```yaml_tag``` is mostly to make things less verbose. I can take a look sometime, ideally the ```load_safe``` method would allow only the ```!``` syntax but not the ```!!python/object:``` one..

pierregodard:
After being recently moved to `requirements-extra.txt`, `sentencepiece>=0.0.6` made it back to `requirements.txt` after this merge.

msperber:
Update: I tried using ```yaml.safe_load```, but it seems it doesn't support the ```!``` syntax, so it would probably be considerable effort to prevent the creation of arbitrary objects via the XNMT config files.

PR #355, opened 2018-04-20
Sentencepiece as optional dependency

msperber:
I've been having trouble installing sentencepiece via pip. I didn't try very hard to solve it, but others might have the same problem so I suggest making it an optional dependency.


neubig:
Good idea!

PR #366, opened 2018-04-27
WER/CER across lines

msperber:
Enable optionally computing WER/CER scores across lines, rather than for each line individually.


neubig:
I think this is fine to merge now, but isn't this a pretty rare use case? I could see merging on a document-by-document level, but merging the whole test set seems to be overkill.

msperber:
I implemented it because I sometimes have relatively poor sentence boundaries. Document-by-document would be even better but would require somehow passing information on document boundaries which I didn't want to implement. I'm perfectly fine keeping this in my local branch though if that seems more appropriate.

neubig:
I'm happy to merge this now, but let's create an issue saying that it'd be nice to be able to preserve information about document boundaries.

PR #370, opened 2018-04-27
Drop masked decoder states

msperber:
This (optionally) drops decoder states when encountering masked regions of a batched target sequence, which can save a good amount of memory (especially with long sequences, such as character sequences). To achieve this, sentences within each minibatch are now sorted by decreasing target length, which I think shouldn't matter in other parts of the code or break anything.

Other than that, I tried to do this in an unintrusive way by writing a ```xnmt.batcher.truncate_batches``` function that takes care of dropping batches and is inserted in the code where necessary.

WIP because this depends on a pending DyNet PR: https://github.com/clab/dynet/pull/1363 that introduces slicing and makes the dropping of batch elements in-placed.


msperber:
Since slicing is merged into DyNet now, I'll mark this as no longer WIP. Would this PR be considered useful? If so, it would be ready to be merged.

Issue #368, opened 2018-04-27
Missing yaml tags/serialization bug?


pmichel31415:
I couldn't load a model containing a ResidualBiRNNBuilder, and I realized that the class didn't have a `yaml_tag` field.

There should be tests in place to prevent this from happening.

So here's the error message:
```python
yaml.constructor.ConstructorError: could not determine a constructor for the tag '!ResidualBiRNNBuilder'
  in "recipes/models/europarl-en-fr.mod", line 106, column 14
```

pmichel31415:
Hmm this is very unfortunate, I'll tag it as major bug

neubig:
Is it fixed by https://github.com/neulab/xnmt/pull/371 ?

(really sorry about this being broken)

pmichel31415:
The BiRNNBuilder part is fixed by #371 

The part about LinearBridge not being serializable isn't. Also IMO there should be automated tests so that this last bug doesn't happen in the future with other modules

msperber:
This is fixed by #372 so I'll close.

PR #371, opened 2018-04-27
assert + fix yaml_tag for all Serializable classes

msperber:
Adresses #368


neubig:
Thanks! This will be useful.

Issue #320, opened 2018-04-11
Transformer Implementation Improvements


neubig:
There are a few improvements that could be made to the transformer implementation:

1. We have not confirmed that it gets competitive performance with the original implementation, nor created a recipe file.
2. It's still a bit slow.
3. It's kind-of separate from the other parts of xnmt, and it'd be better to integrate it by using the standard "translator" class, etc.

*Edit:* Based on @DevSinghSachan 's suggestions, I've summarized things to be fixed, which will probably be fixed in multiple PRs.

Here is a checklist to be finished for speed improvements in training:
- [x] Cleaner padding/masking
- [ ] Avoid loading positional encodings onto the GPU every time by making them parameters
- [x] Remove for loops in multi-head attention
- [ ] Need to make sure DyNet is using batched matrix multiplies: https://github.com/clab/dynet/issues/1354

Here is a checklist of things to be finished to fix decoding:
- [ ] Make the decoding fit in the standard inference classes. currently decoding is done entirely on it's own
- [ ] Avoid re-encoding the source sentence during decoding (probably fixed by the previous one)

Here is a checklist of things that need to be implemented (either in code or in the yaml recipe) for accuracy:
- [ ] Use of different dropout parameters
- [ ] Proper placement of layer norm
- [ ] Initialization of embeddings
- [ ] Make sure that we're using the recommended Adam optimizer
- [ ] Adjust the number of words in the minibatch
- [ ] Make sure that we're using the right BLEU script

DevSinghSachan:
Here are a few of the improvements which I can think of to make the Transformer in xnmt competitive with the one implemented in Tensor2Tensor or the one in Pytorch. I will group the following in 2 categories: speed improvements and model improvements based on my experience developing the model.

Speed Improvements:
(a)  Support of zero-embedding (vector of all zeros) for a padding index in Dynet. Currently, in order to achieve this, we are doing lookup for each word individually so that we can insert tensor of zeros for the padding index. If this feature is supported, we can do batch lookup of all words in a minibatch, and this should improve speed by 1000-2000 wps. I have also tried using external masks for achieving this, but they slow the speed even further.

[...]

PR #372, opened 2018-04-28
Fix serialization issues, add automatic check

msperber:
This addresses #368:
- adds an automatic check and raises an exception when the components written out don't include all of the allocated DyNet parameters. The exception will be raised before the training loop is entered.
- fixes LinearBridge, as well as a couple of other places that had the same problem.

The transformer test currently fails because of this new check, so I marked it as ```@unittest.expectedFailure```. The reason it fails is that it uses a lot of nested classes that are all not Serializable even though they should be, and then allocates Serializable classes at the bottom of the hierarchy without using the ```add_serializable_component``` idiom. It might be worth addressing this later, as part of #320?

### DIFF
--- a/examples/09_programmatic.py
+++ b/examples/09_programmatic.py
@@ -105,6 +105,6 @@ standard_experiment = Experiment(
 )
 
 # run experiment
-standard_experiment(save_fct=lambda: save_to_file(model_file, standard_experiment, ParamManager.param_col))
+standard_experiment(save_fct=lambda: save_to_file(model_file, standard_experiment))
 
 exit()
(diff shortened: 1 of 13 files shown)


msperber:
Strange, we're suddenly getting a segmentation fault when calling loss.backward(). This is not happening when running tests locally, and this PR is doing nothing I can think of that would cause this behavior..

neubig:
Regarding the transformer, I'll be doing some refactoring on this to make it more canonical xnmt style, so don't worry about it for now.

msperber:
Tests are passing now, so this is ready to merge. FYI, the segmentation fault occurred only with the pip version of DyNet and not with the version from the master branch. The workaround was to revert some unimportant and seemingly unrelated changes. I won't investigate further because it seems this is a DyNet bug that has been fixed.

PR #374, opened 2018-04-30
Automatic Bugchecks

### LINKED ISSUE #373: Automatic check for Serializable conventions
The most important checks are in place now, but it would be nice to add checks for the following remaining conventions from http://xnmt.readthedocs.io/en/latest/writing_xnmt_classes.html :
- [ ] uniqueness of class name
- [ ] ```@serializable_init``` decorator being used
- [ ] prevent calling ```my_params()``` after ```__init__()``` was completed
- [ ] prevent passing subcollection to a ```Serializable``` class

msperber:
This fixes #373 by checking automatically that class name are unique, that the ```@serializable_init``` decorator is used, that my_params() is not called after ```__init__()``` was completed, and that ```dy.ParameterCollection``` is not passed as argument to serializable sub-components.

Fixes a good number of places where some of these things were not handled correctly. I don't believe any of those would have actually caused problems, but it's still good to have more consistency.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,9 @@ test/config/logs/
 test/config/models/
 /*.yaml
 examples/preproc/
-/test/config/logs/
-/test/config/models/
+.project
+.pydevproject
+.settings/
+dist/
+docs/_build/
+docs/examples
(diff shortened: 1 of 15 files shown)


neubig:
This is really good, thanks! Feel free to merge when tests pass.

Issue #235, opened 2017-11-10
Multi-reference evaluation


neubig:
It would be nice if we could do evaluation with multiple references, particularly for BLEU.

PR #377, opened 2018-05-01
Multi-reference BLEU scores

msperber:
This solves #235 by adding support for computing BLEU scores against multiple references. Results are exactly the same as I get with multi-bleu.pl.

GLEU, WER, CER evaluators will ignore all but the first reference, so that we can mix multi-reference BLEU and single-reference GLEU, for instance. Other evaluators do not support multiple references and will probably crash if multiple references are specified.

Also includes a little bit of clean up: fast BLEU implementation is moved to its own class and ```evaluate_fast``` is no longer needed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,9 @@ test/config/logs/
 test/config/models/
 /*.yaml
 examples/preproc/
-/test/config/logs/
-/test/config/models/
+.project
+.pydevproject
+.settings/
+dist/
+docs/_build/
+docs/examples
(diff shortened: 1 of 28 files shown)


neubig:
Thanks! I think silently ignoring multiple references is probably a bad idea though. Could we either:
1. Throw an error
2. On a sentence-by-sentence basis, take the highest-scoring reference (I believe this is the standard thing to do for multi-reference evaluation of sentence-level measures)

msperber:
Ok, I went with simply throwing an error for now.

PR #378, opened 2018-05-01
Speed up discrete lexicon creation

neilzxu:
- Using a sparse tensor to directly generate lexicon matrix. It seems to be faster
- Account for unicode when opening lexicon


philip30:
How fast was the speed improvement?

neilzxu:
Previously it was 700 words/sec now it's about 3000 words/sec on a standard translation model w/ vocab of 8000. So around 4x faster?

philip30:
Wow, that's great. I'll merge it. Thank you very much!

PR #379, opened 2018-05-02
Add inference to transformer config

nvog:
The Transformer example config file didn't have an inference object specified, so it was raising an exception during training.


philip30:
I think this is OK but not needed. By default, the SimpleInference will be instantiated I guess? I'll merge it anyway.

PR #380, opened 2018-05-02
Fixed SegmentingSeqTransducer

philip30:
This is all the works on the segmenting transducer. It is working, and I'd want to merge so that I don't have to always merge things manually. This branch is still WIP.


test/config/bow.yaml
```diff
@@ -0,0 +1,39 @@
+# A standard training run, should almost never break
```
msperber: That's probably not a proper comment/description of this test config?

xnmt/decoder.py
```diff
@@ -68,7 +68,8 @@ def __init__(self,
                rnn_layer=bare(UniLSTMSeqTransducer),
                mlp_layer=bare(MLP),
                bridge=bare(CopyBridge),
-               label_smoothing=0.0):
+               label_smoothing=0.0,
+               loss_scaler=None):
```
msperber: Please explain loss_scaler in doc string

xnmt/input.py
```diff
@@ -25,25 +26,38 @@ def get_padded_sent(self, token, pad_len):
 class SimpleSentenceInput(Input):
   """
   A simple sent, represented as a list of tokens
-  
+
   Args:
     words (List[int]): list of integer word ids
-    vocab (Vocab):
+    Annotation (Dict): annotation for the sentence
```
msperber: Could you remove the annotation here, and instead directly pass the information to the component that needs it (maybe using ```!Ref```) if possible? Given the growing number of features in XNMT, I think it is becoming more and more important to keep the basic interface clean. (I think you mentioned before that this would be possible)

xnmt/input_analyzer.py
```diff
@@ -0,0 +1,6 @@
+
+class InputAnalyzer(Serializable):
```
msperber: Can this be removed? If not, it should be commented.

xnmt/translator.py
```diff
@@ -101,7 +102,8 @@ def __init__(self, src_reader, trg_reader, src_embedder=bare(SimpleWordEmbedder)
                encoder=bare(BiLSTMSeqTransducer), attender=bare(MlpAttender),
                trg_embedder=bare(SimpleWordEmbedder), decoder=bare(MlpSoftmaxDecoder),
                inference=bare(SimpleInference), search_strategy=bare(BeamSearch),
-               calc_global_fertility=False, calc_attention_entropy=False):
+               calc_global_fertility=False, global_fertility_weight=None,
+               calc_length_difference=False, length_difference_weight=None):
```
msperber: Please update doc strings.

neubig:
I'll take a look at this, but it seems that it's failing tests at the moment.

philip30:
will be working on this very soon.

philip30:
Hi @neubig and @msperber, this should be ready to review. I have improved the readability and reliability of the code by refactoring and adding the testing. I also added some example of using the SegmentingSequenceTransducer to address #392 .

However, I made some several things that need to be reviewed such as:
- Translator is able to handle several multiple inputs from the decoder. This is addressing #459. Mostly, the default sequence transducer will produce ONE encodings and ONE initial_state. I handle to just simply loop over MULTIPLE encodings and MULTIPLE initial_state when calling the decoder. The loss of the translator is the SUM over all multiple initial_state from the decoder.
- Inference now has start_inference and end_inference event. This probably not a very direct solution for generating a single report file for the whole inference process. The current reporting structure is one file for each input, but in some cases, where for example I want to report my biproduct segmetation, I do not want single line for each file.

Other than that, I also added other some nice features such as:
- DefinedSequence in hyper_parameters.py, we can use this to simulate structural dropout by setting the dropout: DefinedSequence(1.0, 0.9, 0.8, 0.5, 0.25, 0) and basically it can be used to replace any other numbers. The next value is chosen after an epoch is done. I have handled the serialization.
- I pull out the reinforcement learning part to a stand alone PolicyGradient module. I put all of these in xnmt/rl and can be used in the future.

I haven't added the example of learning the segmentation. I will add that after I successfully learn the segmentation from the real dataset.

Thank you!

xnmt/hyper_parameters.py
```diff
@@ -2,91 +2,74 @@
...
+  def update_value(self):
+    return self.value
+
+  # Operators
+  def __lt__(a, b): return a.value < b
+  def __le__(a, b): return a.value <= b
+  def __eq__(a, b): return a.value == b
+  def __ne__(a, b): return a.value != b
+  def __ge__(a, b): return a.value >= b
+  def __gt__(a, b): return a.value > b
```
msperber: This is minor but I think it's possible to simplify this a bit using ```@functools.total_ordering```

xnmt/translator.py
```diff
@@ -130,38 +127,46 @@ def shared_params(self):
             {".attender.state_dim", ".decoder.rnn.hidden_dim"},
             {".trg_embedder.emb_dim", ".decoder.trg_embed_dim"}]
 
+
```
msperber: These changes are similar to things like the compound inputs etc. I’ve also found it difficult to come up with a clean solution. This looks ok in general, although I would suggest using a special data type such as CompoundSeqExpression. This communicates the intention more clearly, as XNMT has “lists” in many dimensions, e.g. list of tokens, list of batch elements, stack of encoder states in a multilayer RNN, etc. It’s easier to keep track of these things without having to read through all of the code by giving it a clear type.

xnmt/hyper_parameters.py
philip30: I took a look at the implementation and I think it's a bit overhead to do so. It is also easier to understand by listing this. So let's keep it this way?

xnmt/reports.py
```diff
@@ -190,47 +190,37 @@ def add_atts(self, attentions, main_content, src, src_str, trg_str, idx):
...
+    words = []
+    start = 0
+    for end in actions:
+      words.append("".join(src[start:end+1]))
+      start = end+1
+    print(" ".join(words), file=self.report_fp)
 
-    self.write_html_tree()
+  @handle_xnmt_event
+  def on_start_inference(self):
```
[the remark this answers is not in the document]
philip30: OK, I will delete the start_inference event.

xnmt/rl/confidence_penalty.py
```diff
@@ -0,0 +1,37 @@
+
+import dynet as dy
+
+from xnmt.persistence import serializable_init, Serializable
+from xnmt.events import register_xnmt_handler, handle_xnmt_event
+
+class ConfidencePenalty(Serializable):
```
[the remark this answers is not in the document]
philip30: I will do so in the next merge request when I have the learning segmentation experiment.

xnmt/priors.py
```diff
@@ -0,0 +1,54 @@
+
+import math
+import numpy as np
+from scipy.stats import poisson
+
+from xnmt.persistence import serializable_init, Serializable
+from xnmt.events import register_xnmt_handler, handle_xnmt_event
+
+class Prior(object):
```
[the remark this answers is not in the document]
philip30: I will document that.

xnmt/hyper_parameters.py
msperber: Sure, I think that's perfectly fine.

philip30:
PTAL @msperber

msperber:
LGTM, thanks for addressing the comments.

philip30:
BTW I think it is a good idea to address the preprocessing issue. I think it is more convenient if we can do preprocessing based on script. We can do that at the next pull request though.

xnmt/priors.py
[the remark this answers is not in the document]
neubig: I still don't see documentation. Also, is this only applicable to the segmenting encoder? If so, then it probably be in the segmenting encoder file.

xnmt/rl/policy_gradient.py
```diff
@@ -0,0 +1,181 @@
...
+from enum import Enum
+from xnmt.events import handle_xnmt_event, register_xnmt_handler
+from xnmt.transform import Linear
+from xnmt.persistence import Ref, bare, Path, Serializable, serializable_init
+from xnmt.rl.eps_greedy import EpsilonGreedy
+from xnmt.constants import EPSILON
+from xnmt.loss import FactoredLossExpr
+from xnmt.param_init import GlorotInitializer, ZeroInitializer
+
+class PolicyGradient(Serializable):
```
neubig: Policy gradient is a training strategy, right, similar to MLE or minimum risk? I'm not sure why this class is separate and not inheriting from the parent class. What is the relationship?

philip30: I think it's a bit different because the previous trainig strategy is a supervised one (calc_loss receive input and reference). This one is unsupervised. Maybe we should later refactor this.

xnmt/priors.py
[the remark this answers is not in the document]
philip30: I think currently it is only being used there. OK I will move it there for this time, but I think this can be used somewhere else later.

xnmt/batcher.py
```diff
@@ -139,8 +141,12 @@ def cmult_by_timestep_expr(self, expr, timestep, inverse=False):
       mask_exp = dy.inputTensor(self.np_arr[:,timestep:timestep+1].transpose(), batched=True)
     return dy.cmult(expr, mask_exp)
 
-  def get_active_one_mask(self):
-    return 1 - self.np_arr
+  @lru_cache(maxsize=1)
```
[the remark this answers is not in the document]
neubig: If so, then that's fine.

xnmt/compound_expr.py
```diff
@@ -0,0 +1,13 @@
+
+class CompoundSeqExpression(object):
```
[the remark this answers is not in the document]
neubig: I don't have a strong preference, either way.

xnmt/batcher.py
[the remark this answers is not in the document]
msperber: I also think the mask was intended to be immutable, and haven't seen any code that modifies it. It might be good to put a short "immutable" comment in the doc string?

xnmt/rl/policy_gradient.py
neubig: OK. I think it's OK to keep this as-is for now. My main concern is if this is floating out in a different place than the other training code then it might be an extra thing to maintain that will get stale. In a future commit let's think about a way to make sure it's tightly integrated with the trainers so we don't have duplicate code.

neubig:
Once you fix the remaining comments I'll be happy to merge.

xnmt/batcher.py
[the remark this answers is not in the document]
philip30: Adding that

script/vocab/make-vocab.py
```diff
@@ -0,0 +1,38 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
philip30: will do so.

xnmt/rl/policy_gradient.py
philip30: Yea, I think once we have the better interface for unsupervised training, I can think of a way to integrate this.

xnmt/compound_expr.py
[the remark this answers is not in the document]
msperber: Yeah, the purpose of this data structure is mainly to convey intended semantics, so more specific sounds good.. how about even something like ```SampledExpressionSeqs``` or ```AlternativeReprExpressionSeqs```?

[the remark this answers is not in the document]
philip30: It's kind of strange for SampledExpressionSeqs because to handle the part of the code that is not a list of expr_seq, I make this data structure with 1 element in it...

philip30:
PTAL @neubig

neubig:
LGTM! Merged

PR #381, opened 2018-05-02
Add the identity as an activation function for the MLP

pmichel31415:


pmichel31415:
FYI this can be useful in the last layer of the decoder when the softwax weights (output projector) are tied with the word embeddings (otherwise this is equivalent to having only one linear layer).

neubig:
Thanks!

PR #386, opened 2018-05-07
Boost </s>

msperber:
This enhances ```BeamSearch``` to support adding a value to the log prob of ```</s>``` for more explicit length control.

### DIFF
--- a/xnmt/length_normalization.py
+++ b/xnmt/length_normalization.py
@@ -46,10 +46,6 @@ class NoNormalization(LengthNormalization, Serializable):
   """
   yaml_tag = '!NoNormalization'
 
-  @serializable_init
-  def __init__(self, boost_eos: Optional[float] = None):
-    super().__init__(boost_eos=boost_eos)
-
   def normalize_completed(self, completed_hyps: Sequence['search_strategy.BeamSearch.Hypothesis'],
                           src_length: Optional[int] = None) -> Sequence[float]:
     return [hyp.score for hyp in completed_hyps]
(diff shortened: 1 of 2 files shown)


msperber:
I hadn't thought of that actually. I renamed ```normalize_partial()``` to ```normalize_partial_topk()``` and added a new ```normalize_partial_all()``` that now handles this. It can't be handled by ```normalize_partial_topk()``` because boosting must occur before selecting the top-k (otherwise in the extreme case of beam size 1, eos-boosting will have no effect at all).

neubig:
I took another look. It seems to me that this `boost_eos` is just one strategy for length normalization and is very similar to `AdditiveNormalization`. Do you think we could merge the two of them together? I'm not a huge fan of having a specific length normalization strategy implemented in the parent class.

msperber:
Yeah it would be nice to find a cleaner way.. In my recent experiments, it was often beneficial to combine ```boost_eos``` and ```PolynomialNormalization``` so I hope to keep that option open. Maybe it would be an idea to support something like a chain of length normalizers?

msperber:
I gave this another try. A chain seemed a bit unintuitive and would have required some more refactoring. Instead, we're now simply passing a Callable to the beam search that, if given, performs some kind of post-processing of the scores for the whole vocabulary.

To combine it with length normalization, one would write:
```
!BeamSearch
  beam: 5
  scores_proc: !EosBooster { boost_val: -3.0 }
  len_norm: !PolynomialNormalization { ... }
```

neubig:
Looks good, thanks!

PR #388, opened 2018-05-09
Improve NormalizerRemovePunct

msperber:
- remove Spanish punctuation
- remove duplicate whitespace


neubig:
This is good, but I wonder if there is a more general way of doing this than just adding Spanish punctuation to the file? Does Python somewhere have access to a list of all punctuation in the language in its localization file? If not, then we can just commit as-is, but if so it might be nice to do it in the more general way.

msperber:
I found a more general way on the web that uses unicode categories and I checked that it works e.g. with Japanese punctuation. I updated the code accordingly.

PR #391, opened 2018-05-10
extended LoadSerialized to support appending to a list

msperber:
Currently, when loading a saved model for evaluation, it's not possible to add a new evaluation task to the already existing ones, instead, the whole list of evaluators has to be re-specified.

This adds an ```append``` syntax that allows doing, for example:

```
!LoadSerialized:
  path: ...
  overwrite:
  - path: evaluate.append
     val: ...
```

Also includes some cosmetic improvements to the doc strings.

### DIFF
--- a/xnmt/persistence.py
+++ b/xnmt/persistence.py
@@ -4,14 +4,14 @@ same YAML file format.
 
 The main objects to be aware of are:
 
-* Serializable: must be subclassed by all components that are specified in a YAML file.
-* Ref: a reference that points somewhere in the object hierarchy, for both convenience and to realize parameter sharing.
-* YamlPreloader: pre-loads YAML contents so that some infrastructure can be set up, but does not initialize components.
-* initialize_if_needed, initialize_object: initialize a preloaded YAML tree, taking care of resolving references etc.
-* save_to_file: saves a YAML file along with registered DyNet parameters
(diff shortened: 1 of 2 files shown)


neubig:
Hmm, this seems a bit confusing because the syntax for appending something is the same as the syntax for specifying a path. How about instead of this, we do 

    - path: xxx
       append_val: xxx

or something?

msperber:
That sounds good, done.

xnmt/persistence.py
```diff
@@ -535,13 +536,17 @@ def set_child_serializable(node, name, val):
 
 @set_child.register(list)
 def set_child_list(node, name, val):
+  if name == "append": name = len(node)
```
neubig: What happens if someone accidentally creates a member named "append" in a class they're implementing? Would this cause strange impossible-to-trace errors in the serialization process? From my reading of the code it seems so. If so, I'd suggest naming "append" something that people are less likely to use ("__append"?) and throw an error if you detect that someone has accidentally used this name.

msperber: The ```@set_child.register(list)``` makes sure that this is only called when ```node``` is a subtype of ```list``` which accepts only integers as keys, so I think this should be safe.

PR #397, opened 2018-05-14
add docopt to requirements.txt

msperber:
script/code/conv_checkpoints_to_model.py uses docopt, so I added it to the requirements.


neubig:
Should this be in `requirements-extra.txt`?

msperber:
I moved it, will merge

Issue #398, opened 2018-05-14
Broken Pretrained Embeddings Option


pierregodard:
I would like to use some pretrained embeddings for a low-resource setting, but as indicated in `pretrained_embeddings.yaml`, this option (PR [#197](https://github.com/neulab/xnmt/pull/197)) is broken, producing the following error:
```
File "stringsource", line 2, in _dynet.LookupParameters.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
```
I don't feel quite equipped to start understanding what needs to be repaired. Could you (@msperber?) tell me how hard this would be?
Thank you!

msperber:
The error indicates that we're trying to write Cython object to YAML which does not work. I'm not sure why that happens. Could you try adding ```Serializable``` as second base class to ```PretrainedSimpleWordEmbedder```? It seems like this should be inherited from ```SimpleWordEmbedder```, but I remember having an issue like this before.

msperber:
This is probably unrelated, but ```PretrainedSimpleWordEmbedder``` is also missing ```@register_xnmt_handler``` which would lead to problems when using dropout or weight noise.

pierregodard:
Adding `Serializable` to `PretrainedSimpleWordEmbedder` gets the example running again! Thank you so much @msperber. (I also added `@register_xnmt_handler` to the `__init__` method afterwards, but this was not causing the issue.)
I will experiment with this pretrained option today, but do you want me to create a mini-PR, or this is more distracting than fixing it yourself in the master? Thank you again for being so reactive.

pierregodard:
Will do. Thank you!

PR #400, opened 2018-05-15
Fix broken pretrained embeddings option

pierregodard:
Fixes [#398](https://github.com/neulab/xnmt/issues/398).


msperber:
Thanks!

PR #401, opened 2018-05-17
SentenceFilterLength: fix min sentence length and max/min index

gmwe:
For multiple min/ max values (min_0, ..) fixed the min assignment and the assignment in general.


msperber:
Thanks!

PR #403, opened 2018-05-22
fix dev loss word count is taken from the first dev task

gmwe:
The dev word count was previously taken from the last dev task, but the score from the frist task.
Now the word count is taken from the first task as well.


msperber:
Nice catch, thanks!

PR #406, opened 2018-05-23
Generalize loss tracker

msperber:
Currently, having anything but ```SimpleSentenceInput``` as targets fails because the loss tracker's ```count_trg_words``` (which is needed to compute the per-word loss) assumes these as targets.

This is a fix, although I'm wondering if we could simplify it even further. As of the current PR, everything except </s> is counted as a word in case of ```SimpleSentenceInput``` (consistently with current implementation on master), otherwise the number of unmasked items in each batch element is counted (new). This is not the same because for ```SimpleSentenceInput``` the first </s> is still unmasked, only follow-up tokens are masked. I'm not sure if this is an important distinction, if it is not it would be better to simplify this further by removing the ```SimpleSentenceInput``` special case.


neubig:
Could we move the counting of target words into the `Input` class itself? This seems like it would allow things to be more flexible if we have other data structures on the target side.

msperber:
That's certainly a better solution, done.

PR #407, opened 2018-05-24
Preproc regex filter

gmwe:
Added a preprocessing filter that can filter sentences by regular expression.

Example usage:
```
    - !PreprocFilter
      in_files:
      - examples/data/head.ja
      - examples/data/head.en
      out_files:
      - examples/output/head.filter.ja
      - examples/output/head.filter.en
      specs:
      - type: matching-regex
        regex_trg: "you"
```


neubig:
Thanks! For a later commit, we may want to change the specification of the filter to the recent style of doing `!SentenceFiltererMatchingRegex:` instead of `type: matching-regex`.

PR #411, opened 2018-05-28
Configure how to combine batched losses

msperber:
Choices are ```sum``` and ```avg```.

It took me some time to understand the semantics of the loss builder, so I tried documenting and renaming things to make it easier to understand. Maybe @philip30 could let me know in case I misunderstood anything.


neubig:
LGTM, can be merged after resolving conflicts.

PR #412, opened 2018-05-30
[WIP] Sequence Classifier

msperber:
This PR adds a sequence classifier model (src embedder + encoder + MLP softmax over mean of encoder states). Includes:
- ```SequenceClassifier``` in classifier.py
- add example config and config for integration tests
- some restructuring for inference:
  - add a new ```Inference``` base class
  - rename ```SimpleInference``` -> ```SequenceInference``` (and splitting the large main method into smaller methods for better readability)
  - add new ```ClassifierInference```
- some cleanup and fixes to outputs

UPDATE: I'll mark this as WIP as I'm working on extending/refactoring this some more.


msperber:
One point up for discussion would be whether the loss calculator should be part of DefaultTranslator or of the training strategy. It's currently part of the training strategy, and the sequence classifier simply ignores it because the currently implemented loss calculators are specific to attentional seq2seq models. It seems like adding unnecessary complexity to introduce a new loss calculator just for the classifier model, so we could ```assert loss_calculator is None``` or something similar.

msperber:
I'll close this for now and will send a new extended PR at some later point.

PR #413, opened 2018-06-04
update_every

msperber:
This PR implements a new ```update_every``` option for all training regimens that simply skips the call to ```trainer.update()``` for an appropriate number of minibatches in case ```update_every > 1```. Also includes some improved type annotations for SimpleTrainingRegimen.


msperber:
I'm closing this in favor of #414.

PR #414, opened 2018-06-07
Optimization features

msperber:
This replaces #413 and adds 3 features to the optimizers that can help stabilize training:
- ```update_every```: simulated large-batch training
- ```skip_noisy```: keep track of a moving average and a moving standard deviation of the log of the gradient norm values, and abort a step if the norm of the gradient exceeds four standard deviations of the moving average. Reference: https://arxiv.org/pdf/1804.09849.pdf

UPDATE: ```clip_norm``` is already the default DyNet behavior and is now removed from this PR.

### DIFF
--- a/docs/save_file_format.rst
+++ b/docs/save_file_format.rst
@@ -1,4 +1,4 @@
-.. _sec-programming-style:
+.. _save-file-format:
 
 Save File Format
 ================
--- a/examples/12_multi_task.yaml
+++ b/examples/12_multi_task.yaml
@@ -11,12 +11,13 @@ exp1-multi_task: !Experiment
     default_layer_dim: 64
   train: !SameBatchMultiTaskTrainingRegimen
     trainer: !AdamTrainer {}
+    n_task_steps: [2,1]
     tasks:
     - !SimpleTrainingTask # first task is the main task: it will control early stopping, learning rate schedule, model checkpoints, ..
       name: first_task
       run_for_epochs: 6
       batcher: !SrcBatcher
-        batch_size: 6 # batch size is twice as big as for task 2, which will give task 1 more impact during training
+        batch_size: 6
       src_file: examples/data/head.ja
       trg_file: examples/data/head.en
       model: !DefaultTranslator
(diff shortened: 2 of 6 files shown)


pmichel31415:
Correct me if I'm wrong but I think dynet already does gradient clipping this way.

msperber:
Thanks for the hint! I guess I had misunderstood the documentation, but looking at the DyNet code suggests you are right.

msperber:
Update: I realized that ```update_every``` should be part of the training regimen, not the optimizer, after all. This allows implementing a custom strategy for the alternating-batch multi-task regimen which will draw from the same task multiple times until an update is performed. This seems the expected behavior here, otherwise one would have probably resorted to the same-batch variant. I've updated the code accordingly.

neubig:
Thanks, this looks good!

Actually, I don't know if the current behavior for AlternatingBatchTrainingRegimen is what I'd expect though. The reason why I wasn't using "SameBatch" was because I wanted a fixed ratio of data from one task to another task, for example 4-to-1, but still wanted to maintain the computational advantage of not having to make one of my batch sizes four times smaller than the other. It'd be nice if, for example, it was possible to sample 4 batches individually according to this 4-1 ratio, then update after calculating the gradients for all of them.

I do agree that the "update_every" should be in the TrainingStrategy though.

msperber:
Yeah, I can see that..

Currently, if one uses "AlternatingBatch" with a 4-to-1 ratio and then set ```update_every``` to say 2, it would still draw a task according to this ratio, then draw two minibatches for it, then update, etc. So the ratio is still supported, but gradients are not accumulated across tasks.

For the strategy you mentioned, how about adding the capability to ```SameBatchTrainingRegimen```, e.g. via a ```steps_per_task``` field that takes a list of numbers of minibatches to draw for each task? Passing ```[2,2]``` would have the same effect as setting ```update_every=2```, but to realize your example you would pass ```[4,1]``` and set ```update_every=1```.

neubig:
I actually can't think of any reason why you'd prefer to take N consecutive samples from the same task in the alternating task batcher though. It seems like this could only increase instability in the training process and has very few advantages. Can you think of any?

neubig:
Yeah, I think having this be configurable would be a good idea.

msperber:
Okay, I implemented this so that for the same-batch strategy we can now configure task weights directly, and for the alternating-batch strategy we can do update_every either repeatedly for the same task or across randomly drawn tasks. Should be ready to merge from my side.

neubig:
Thanks!

PR #416, opened 2018-06-11
Self-Attentional Acoustic Model

msperber:
This is the main code for the "Self-Attentional Acoustic Models" Interspeech paper: https://arxiv.org/abs/1803.09519

I created a new ```xnmt.specialized_encoders``` package where I put both this and some of the retrieval-related code.

Note: this PR includes only the "interleaved" variant. The "stacked" variant requires some more cleanup and I'll send it in a later PR.


neubig:
Cool! I'm probably not going to have time to read this closely, but I'm happy to have you merge it, given that it's going to go into the "specialized encoders" part. It'd also be nice to have a recipe if possible.

msperber:
I'm planning on preparing some config files as well, but would do that in a later commit (a test config file is already included in this PR).

neubig:
Sure, go ahead and merge this then!

Issue #417, opened 2018-06-11
Better doc about adding classes


neubig:
It would be nice to have an example of how to create a new xnmt class, maybe with a very basic example in the doc here:

> http://xnmt.readthedocs.io/en/latest/writing_xnmt_classes.html

For example, some things that's are not written are where to look for an example that can be copy-pasted, and also where the class name needs to be added to it can be recognized in config files.

PR #418, opened 2018-06-11
Stand-alone multi-headed self attention encoder

neubig:
We now have a couple of specialized encoders that implement self-attention, including the transformer encoder and now the self attentional acoustic models. However, this somewhat goes against the idea of xnmt being a toolkit where you can easily mix and match methods, so I'd like to move back towards the bigger vision of xnmt allowing you to mix and match methods.

This is a first step: implementation of multi-headed attention as a stand-alone encoder. Because of the following issue, I'm not sure where is the canonical place to add this to the codebase so it can be referenced from the yaml files, but other than that I think it should be OK. It still needs testing though...
https://github.com/neulab/xnmt/issues/417


xnmt/self_attention.py
```diff
@@ -0,0 +1,87 @@
...
+
+class MultiHeadAttentionSeqTransducer(SeqTransducer, Serializable):
+  """
+  This implements the Multi-headed attention layer of "Attention is All You Need":
+  https://papers.nips.cc/paper/7181-attention-is-all-you-need.pdf
+
+  Args:
+
+  """
+  yaml_tag = '!MultiHeadAttentionSeqTransducer'
```
msperber: To support referencing from config files, the module would need to be imported under xnmt/__init__.py

xnmt/self_attention.py
```diff
@@ -0,0 +1,87 @@
...
+
+  @register_xnmt_handler
+  @serializable_init
+  def __init__(self,
+               input_dim=Ref("exp_global.default_layer_dim"),
+               param_init=Ref("exp_global.param_init", default=bare(GlorotInitializer)),
+               bias_init=Ref("exp_global.bias_init", default=bare(ZeroInitializer)),
+               num_heads=8,
+               yaml_path=None):
```
msperber: yaml_path seems unused

xnmt/self_attention.py
```diff
@@ -0,0 +1,87 @@
...
+    param_collection = ParamManager.my_params(self)
+
+    self.input_dim = input_dim
+    self.num_heads = num_heads
+    self.head_dim = input_dim // num_heads
+
+    # self.Wq, self.Wk, self.Wv, self.Wo = [param_collection.add_parameters(dim=(input_dim, input_dim), init=param_init) for _ in range(4)]
+    # self.bq, self.bk, self.bv, self.bo = [param_collection.add_parameters(dim=(input_dim), init=bias_init) for _ in range(4)]
+    self.Wq, self.Wk, self.Wv, self.Wo = [param_collection.add_parameters(dim=(input_dim, input_dim)) for _ in range(4)]
+    self.bq, self.bk, self.bv, self.bo = [param_collection.add_parameters(dim=(1, input_dim)) for _ in range(4)]
```
msperber: These should probably use the passed parameter initializers, similar to e.g. https://github.com/neulab/xnmt/blob/master/xnmt/linear.py#L29

xnmt/self_attention.py
neubig: Removed.

xnmt/self_attention.py
neubig: Fixed, thanks!

neubig:
Tests seem to be failing on an unrelated test not connected to the changes here. I'm not sure what's going on...

neubig:
OK, the failing tests were simply because the Travis version of DyNet was using an older version of DyNet than I was (because we haven't updated the version of DyNet in a while). More worryingly, the forced decoding tests aren't dying, so I'll create an issue for this.

Issue #423, opened 2018-06-14
Forced Decoding Tests are Failing


neubig:
Running the following test

    python test/test_run.py TestRunningConfig.test_forced

Results in the following errors:

    => Running forced
    > use randomly initialized DyNet weights of all components
      DyNet param count: 98053
    > Training
    The dy.parameter(...) call is now DEPRECATED.
            There is no longer need to explicitly add parameters to the computation graph.
            Any used parameter will be added automatically.
    [forced] Epoch 1.0000: train_loss/word=4.709079 (words=91, words/sec=3989.86, time=0-00:00:00)
    > Checkpoint
    [forced] Epoch 1.0000 dev Loss: 4.614 (words=91, time=0-00:00:00)
    [forced]              checkpoint took 0-00:00:00
      best dev score, writing out model
    [forced] Epoch 2.0000: train_loss/word=4.623288 (words=91, words/sec=6980.66, time=0-00:00:00)
    > Checkpoint
    [forced] Epoch 2.0000 dev Loss: 4.516 (words=91, time=0-00:00:00)
    [forced]              checkpoint took 0-00:00:00
      best dev score, writing out model
    reverting learned weights to best checkpoint..
    > Performing final evaluation
    ERROR: Forced decoding score -35.69606590270996 and loss -56.88883972167969 do not match at sentence 0
    ERROR: Forced decoding score -41.25044846534729 and loss -58.03459167480469 do not match at sentence 1
    ERROR: Forced decoding score -28.295342445373535 and loss -44.06516647338867 do not match at sentence 2
    ERROR: Forced decoding score -44.06516671180725 and loss -44.985591888427734 do not match at sentence 3
    ERROR: Forced decoding score -37.07419157028198 and loss -41.250450134277344 do not match at sentence 4
    ERROR: Forced decoding score -56.88884139060974 and loss -35.69606399536133 do not match at sentence 5
    ERROR: Forced decoding score -28.245083808898926 and loss -37.074188232421875 do not match at sentence 6
    ERROR: Forced decoding score -44.98559355735779 and loss -36.42924118041992 do not match at sentence 7
    ERROR: Forced decoding score -36.42923879623413 and loss -28.29534339904785 do not match at sentence 8
    ERROR: Forced decoding score -58.03459358215332 and loss -28.24508285522461 do not match at sentence 9

PR #424, opened 2018-06-15
Fix forced decoding error

neubig:
This is a PR to fix https://github.com/neulab/xnmt/issues/423

The first commit is simply to ensure that tests fail until this error is fixed.

### DIFF
--- a/xnmt/batcher.py
+++ b/xnmt/batcher.py
@@ -1,4 +1,4 @@
-from typing import Optional, Sequence, Tuple, Union
+from typing import Any, Callable, Optional, Sequence, Tuple, Union
 import math
 import random
 import numpy as np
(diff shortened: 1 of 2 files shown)


neubig:
I'm pretty sure I know what the problem is. The code in https://github.com/neulab/xnmt/pull/370 is sorting hypotheses by the target length internally, which means that the results of the loss function are not in their original order any more. The forced decoding debug option assumes that it gets the losses in order so it can compare them with the score it gets from decoding.

Looking at this more closely, I think changing the order of losses by default is a bad idea, because this might also break, for example, xnmt's scoring functionality. We should do either or both of:
1. Have sorting be off by default.
2. Even when sorting is on, before returning any results, make sure the outputs are re-arranged to the original order.

@msperber, would you be able to take a look at this?

msperber:
- When using a batcher to create a single batch, I think you are definitely right that the order should not be changed by default (I had wrongly assumed that the corresponding method was only used from within the module, and have updated the interface to make this clearer).
- When packing batches I would argue that it's ok to change the within-batch order because most batcher's don't make any guarantees for the within-batch order, and things are always consistent because the source- and target-side are sorted together. Sorting merely changes behavior from "unspecified" to "sorted".
- The one exception that I had missed was the ```InOrderBatcher``` which does give guarantees on within-batch order.

My fix thus leaves the trg-length sorting intact for all batchers except the ```InOrderBatcher```. If that seems reasonable, this would be ready for merge.

PR #425, opened 2018-06-15
Fix list item ref-sharing

msperber:
This fixes a problem that prevented reference-sharing of components that are list / dictionary items (e.g. sharing individual modules of the modular seq transducer).


neubig:
Thanks!

PR #426, opened 2018-06-15
Fixed size attention

msperber:
This implements the basic fixed-size memory model according to Britz et. al 2017: ﻿Efficient Attention using a Fixed-Size Memory Representation; https://arxiv.org/abs/1707.00110

It's implemented as a sequence transducer and can, for example, be used by specifying a modular transducer with an LSTM module followed by this new module. I've tested it on a character translation task and got results in line with the paper.

### DIFF
--- a/xnmt/fixed_size_att.py
+++ b/xnmt/fixed_size_att.py
@@ -1,0 +1,55 @@
+from typing import Optional
+
+import numpy as np
+import dynet as dy
+
+from xnmt.persistence import Serializable, serializable_init, bare, Ref
+from xnmt import expression_sequence, param_collection, transducer
+import xnmt.param_init
+
+
+class FixedSizeAttSeqTransducer(transducer.SeqTransducer, Serializable):
+  """
+  A fixed-size attention-based representation of a sequence.
+
+  This implements the basic fixed-size memory model according to Britz et. al 2017: ﻿Efficient Attention using a
+  Fixed-Size Memory Representation; https://arxiv.org/abs/1707.00110
+
+  Args:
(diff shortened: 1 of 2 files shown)


xnmt/fixed_size_att.py
```diff
@@ -0,0 +1,55 @@
...
+               param_init: xnmt.param_init.ParamInitializer = Ref("exp_global.param_init",
+                                                                  default=bare(xnmt.param_init.GlorotInitializer))):
+    subcol = param_collection.ParamManager.my_params(self)
+    self.output_len = output_len
+    self.W = subcol.add_parameters(dim=(hidden_dim, output_len),
+                                   init=param_init.initializer((hidden_dim, output_len)))
+    self.pos_enc_max = pos_enc_max
+
+  def __call__(self, x):
+    x_T = dy.transpose(x.as_tensor())
```
neubig: You could use `x.as_transposed_tensor()` which might save memory in some cases.

xnmt/fixed_size_att.py
```diff
@@ -0,0 +1,55 @@
...
+    scores = x_T * dy.parameter(self.W)
+    if x.mask is not None:
+      scores = x.mask.add_to_tensor_expr(scores, multiplicator=-100.0, time_first=True)
+    if self.pos_enc_max:
+      seq_len = x_T.dim()[0][0]
+      pos_enc = np.zeros((seq_len, self.output_len))
+      for k in range(self.output_len):
```
neubig: This seems quite wasteful to calculate every time. Is there a way it can be cached?

msperber:
Thanks for the hints, I've reflected both.

PR #427, opened 2018-06-15
Unicode tokenizer

msperber:
Implements the reversible, language-independent tokenizer from the appendix of https://arxiv.org/pdf/1804.08205

While at it, I also made some places a bit more YAML-like. I did not change the filterers because their initializer arguments are a bit more complicated and might require **kwargs syntax which currently is not supported by the XNMT serializer (although it would be nice to have..).

### DIFF
--- a/examples/05_preproc.yaml
+++ b/examples/05_preproc.yaml
@@ -4,13 +4,8 @@ preproc: !Experiment
   exp_global: !ExpGlobal
     # define some named strings that can be used throughout the experiment config:
     placeholders:
-      # Specify <SENTENCEPIECE_SRC_PATH> with src/ directory of sentencepiece repository
-      # The spm_* executables should be located at <SENTENCEPIECE_SRC_PATH>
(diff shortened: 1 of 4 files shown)


examples/05_preproc.yaml
```diff
@@ -32,16 +27,18 @@ preproc: !Experiment
...
-          train_files:
-           - '{DATA_IN}/train.ja'
-           - '{DATA_IN}/train.en'
-          vocab_size: 8000
-          model_type: unigram
-          model_prefix: '{DATA_OUT}/sentpiece'
+        - !UnicodeTokenizer {}
+        # uncomment if you have installed the corresponding third-party tool:
+#        - !SentencepieceTokenizer
+#          # Replace <SENTENCEPIECE_SRC_PATH> with src/ directory of sentencepiece repository
```
neubig: This is old information. The sentencepiece tokenizer now uses the Python libraries.

msperber: Thanks, I removed it.

PR #428, opened 2018-06-16
subword regularization

cindyxinyiwang:
subword regularization by sampling


xnmt/input_reader.py
```diff
@@ -122,37 +124,46 @@ def count_words(self, trg_words):
...
+    self.subword_model.Load(model_file)
+    self.vocab = vocab
+    self.include_vocab_reference = include_vocab_reference
+    if vocab is not None:
+      self.vocab.freeze()
+      self.vocab.set_unk(Vocab.UNK_STR)
+
+  def read_sent(self, sentence, filter_ids=None):
+    vocab_reference = self.vocab if self.include_vocab_reference else None
+    words = self.subword_model.SampleEncode(sentence.strip(), -1, 0.1)
```
neubig: `-1` and `0.1` are parameters, right? Could you make these configurable by adding them to the constructor?

xnmt/training_task.py
```diff
@@ -213,7 +213,9 @@ def advance_epoch(self):
         self._augment_data_initial()
       else:
         self._augment_data_next_epoch()
-    if self.training_state.epoch_num==0 or self.sample_train_sents:
+    if self.training_state.epoch_num==0 or self.sample_train_sents or \
+      isinstance(self.model.src_reader, input_reader.SubwordSampleTextReader) \
+      or isinstance(self.model.trg_reader, input_reader.SubwordSampleTextReader):
```
msperber: I'm wondering if we should make this more general, i.e. add a ```is_deterministic()``` or ```needs_reload()``` to input readers and using that instead of checking the type. This would make the code directly ready for future input readers that need reloading between epochs (data augmentation, noisy inputs, ..).

neubig: Very good idea. I agree.

neubig:
Thanks!

Issue #382, opened 2018-05-02
Better Separation of "Softmax"


neubig:
I think that it'd be nice to have better separation of the softmax from the decoder class. Currently we have `MlpSoftmaxDecoder`, but I think it'd be good to have `StandardDecoder` and then maybe something like:

    ModularSoftmax:
    - Mlp:
      ...
    - StandardSoftmax: {}

This would also allow for a more clean integration of stuff like the lexicon bias, with

    ModularSoftmax:
    - Mlp:
      ...
    - LexiconBiasedSoftmax:
      ...

neubig:
CC: @philip30 @pmichel31415 because you've worked on it before, but I think any interested person could do this.

PR #429, opened 2018-06-18
Generalized translator and inference

msperber:
The current design of translator, decoder, and inference classes is quite specific to RNN-based attentional encoder-decoder models which makes it difficult to re-use much of the code when implementing alternatives. One place this problem becomes evident is the current diverge between transformer- and RNN-based models.

This PR does some preliminary refactoring, but is mainly meant to initiate discussions about how to properly design the interfaces.

Description of changes so far:
- improved interfaces: add a ```Inference``` base class, add some missing content to ```Decoder``` base class, add type annotations in some places
- rename ```SimpleInference``` to ```SequenceInference```
- add a sequence labeler and classifier as two simple examples of models that need a different inference strategy and might be instructive for designing better interfaces.
- some other clean up in related code

Things to do:

- [X] better design of inference classes: the current PR implement ```ClassifierInference``` and ```SequenceInference``` which both derive from the ```Inference``` base class. A few problems I noticed so far are that (1) ```SequenceInference``` assumes use of a search strategy which does not make sense for non-autoregressive models (e.g. the sequence labeler) and (2) having a separate ```ClassifierInference``` is probably reasonable, but some things like forced decoding apply here as well and should be reflected in the base class.
- [X] MLELoss: this could probably be generalized to handle non-attentional and potentially non-autoregressive models; otherwise should probably be renamed to communicate its intended use.
- [X] better design of MLP class: it's doing some checks of its place in the component hierarchy that don't generalize well
- [X] Clearer interface for translators
- [X] new sequence labeler and classifier models to verify appropriateness of new interfaces
- [x] separated softmax and projection (#440 )
- [x] fix example configs
- [x] add entries to API doc


neubig:
Thanks! I'm happy to have the code separated better. A few thoughts:

* `SequenceInference` might be `AutoRegressiveInference`?
* It might be a good idea to either try implementing a sequence labeler (just BiLSTM, we could do something like CRF later) or a self-attentional decoder to try to make sure the interface works for these. This could either be during this PR, or after.
* Agree about MLELoss.

docs/experiment_config_files.rst
```diff
@@ -85,7 +85,7 @@ This specifies the model architecture. An typical example looks like this
     decoder: !MlpSoftmaxDecoder
```
neubig: For a while I've disliked the name `MlpSoftmaxDecoder`. This seems to be a good time to change it, because we're changing lots of things. This would also probably allow us to kill this issue:
> https://github.com/neulab/xnmt/issues/382

I'd suggest maybe for the decoder having `mlp:` and `softmax:` specified separately, where the MLP layer is just in charge of transforming one vector to another, and the softmax calculates the probabilities, n-best, etc.

Also, I think it'd be a lot better if we could separate out the attention from this. These are all independent things, and having them in one big package seems to be limiting. Or alternatively, we could have one big package (`AttentionalOutput`?) that has each of these elements as a component.

neubig:
@msperber Are you currently looking at my comment about separating the MLP and softmax classes? If not, I can take a look.

msperber:
@neubig No, I'm not working on this currently, so feel free to go ahead.

msperber:
Alright, this should be ready.

neubig:
I made a few changes and LGTM! But quite bizarrely, it seems that tests are now not passing. The only changes I made should have no possibility of causing this effect (just deleted an unused yaml file and changed some documentation), so maybe it's due to a difference in the Travis environment?

neubig:
P.S. also, tests are passing on my machine.

msperber:
It seems that this is due to an upgrade in pyyaml that introduced some major changes related to making loading safe which I don't fully understand and don't seem to be well documented. I'm downgrading to the previous version for now.

philip30:
Hey @msperber can you also refactor my code (LexiconDecoder)? I don't have time to read the whole refactoring and my code does not work anymore. I think simply just because you are not using it, you can't just remove it from master branch.

If you have any question about it, you can always ask me.

Thank you.

neubig:
@philip30 I did this actually. The reason why is because it shouldn't be a decoder, but rather a softmax. I think it should be implemented from scratch. Perhaps we should make an issue though.

philip30:
@neubig Alright, I'll make an issue!

Issue #430, opened 2018-06-18
Switch train/test behavior of InputReaders


neubig:
InputReaders should know whether they're processing the training or testing sets, in case such as subword regularization https://github.com/neulab/xnmt/pull/428

neubig:
Fixed by https://github.com/neulab/xnmt/pull/431

PR #431, opened 2018-06-18
Made train/test time behavior of input readers different

neubig:
This makes it possible to moderate the train/test time behavior of InputReaders differently, and shows an example with sampling from a SentencePiece model:
https://github.com/neulab/xnmt/issues/430

There are some other changes:
* SentencePieceInputReader can now be used without sampling at all, if you want to just apply a model before-hand.
* Several bugs in the example configuration for subword regularization are fixed (added "join-pieces", made vocab reading correct, etc.)
* SentencePiece functions are switched to "EncodeAsPieces".

### DIFF
--- a/xnmt/training_task.py
+++ b/xnmt/training_task.py
@@ -160,12 +160,14 @@ class SimpleTrainingTask(TrainingTask, Serializable):
       if self.training_state.epoch_num > 0:
         logger.info('using reloaded data')
       # reload the data   
+      self.model.src_reader.train = self.model.trg_reader.train = True
       self.src_data, self.trg_data, self.src_batches, self.trg_batches = \
           input_reader.read_parallel_corpus(self.model.src_reader, self.model.trg_reader,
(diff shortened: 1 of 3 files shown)


xnmt/training_task.py
```diff
@@ -160,12 +160,14 @@ def _augment_data_next_epoch(self):
       if self.training_state.epoch_num > 0:
         logger.info('using reloaded data')
       # reload the data   
+      self.model.src_reader.train = self.model.trg_reader.train = True
```
msperber: In other components, different train/test behavior is achieved by listening to the ```on_set_train()``` event. It might keep things cleaner (and preserve consistency among training regimens) to use that here also unless there is a reason not to? From taking a quick look I think one would have to add a ```set_train(True)``` trigger to ```advance_epoch()```, but otherwise events are probably already triggered at the proper times.

neubig: Done

PR #432, opened 2018-06-18
Printing of debugging on crashes

neubig:
This will help at least give some hints where things are crashing. We can probably make the logging more extensive later by having each Input print info about its representation (which sentence number is it, etc.)

### DIFF
--- a/xnmt/eval_task.py
+++ b/xnmt/eval_task.py
@@ -72,16 +72,17 @@ class LossEvalTask(EvalTask, Serializable):
     loss_val = FactoredLossVal()
     ref_words_cnt = 0
     for src, trg in zip(self.src_batches, self.ref_batches):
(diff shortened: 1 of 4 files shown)


msperber:
Yeah, that's a good idea!

One suggestion would be to make this into a syntax like:
```
with report_on_exception({"src":src,"trg":trg}):
  ...
```
where the with statement contains the try/except on logging. That would make it easy to reuse in other training regimens also.

neubig:
Thanks for the suggestion! I'm actually not familiar with this usage of the "with" syntax in Python though. Would you mind fleshing out the example a little more? (either in the comments or by directly pushing to the branch)

msperber:
Sure, I've pushed something along these lines as a suggestion.

PR #440, opened 2018-06-25
Separated softmax and projection

neubig:
This starts separating out the softmax from the MLP. Basically before, to calculate the log probability over the vocabulary given input `x`, there were two procedures:

* `mlp.MLP`: `s = W_2 * nonlinearity(W_1*x+b_1) + b_2`
* Spread out over `MlpSoftmaxDecoder`, `SeqLabeler`, `SeqClassifier`: `dy.softmax(s)` or `dy.pickneglogsoftmax(s)`

Now, there are two new files `transform.py` (which subsumes the previous `linear.py`), and `scorer.py`

* `transform.NonLinear`: `h = nonlinearity(W_1*x+b_1)`
* `scorer.Softmax`: `s = W_2 * h + b_2`, plus additionally taking the softmax, log softmax, or loss calculation.

I've temporarily removed all references to `MLP`, because it's messy. This might be re-added to `xnmt.transform` later.

@msperber Does this design look OK? If so, I'll go ahead and clean up all the loose ends.


msperber:
Yes, looks okay to me.

neubig:
Great, will continue then!

neubig:
@msperber OK, this should be ready to review now. All tests should pass and most of the examples should work, but some of the examples seemed to be broken previously.

neubig:
One thing: due to https://github.com/neulab/xnmt/issues/443 I still hadn't figured out how to fix the partial population bug, but I'll fix that now. I think it's just a small thing, so the rest of the code is ready for review.

neubig:
Hmm, actually I'm stuck on this... Here is an example:

    xnmt test/config/score.yaml

which yields

    => Running exp2-score
    > populated a subset of DyNet weights from given data files: ['SimpleWordEmbedder.229c4ec2', 'UniLSTMSeqTransducer.11cd2b5b', 'UniLSTMSeqTransducer.e31a7233', 'UniLSTMSeqTransducer.633dee13', 'UniLSTMSeqTransducer.4b8a99d4', 'MlpAttender.7ed9bd8d', 'SimpleWordEmbedder.1c0ab30d', 'UniLSTMSeqTransducer.1e701163', 'Linear.c07a2a47', 'Softmax.db944f1e', 'AutoRegressiveDecoder.4ca9c06c'].
      Did not populate {'AuxNonLinear.d6fbef7e'}.
      If partial population was not intended, likely the unpopulated component or its owner   does not adhere to the Serializable protocol correctly, see documentation:
       http://xnmt.readthedocs.io/en/latest/writing_xnmt_classes.html#using-serializable-subcomponents

However, looking at the file, I'm passing this in as an initializer argument, which seems to indicate that it should be OK: https://github.com/neulab/xnmt/blob/22b5b2e0f2d601d7d60db99aa839ca3124a5e4ea/xnmt/decoder.py#L75

msperber:
Okay, it turns out this was a bug in the persistence module. I pushed a fix to this branch, hope everything is working now.

test/config/assemble.yaml
```diff
@@ -1,4 +1,6 @@
 exp1-pretrain1: !Experiment
+  exp_global: !ExpGlobal
+    model_file: 'test/tmp/{EXP}.mod'
```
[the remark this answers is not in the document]
neubig: I added this because things were breaking, so maybe there's a bug somewhere?

neubig:
I added a little bit of doc, but tell me if there's anything else that needs to be updated.

For the sequence labeler, would you be able to fix this quickly? I would have been happy to, but I didn't understand the logic and didn't want to break anything. Alternatively, we could leave this disabeled for another PR, as I don't think anyone is using it now.

msperber:
Okay, I think we can merge this into the generalize-translator branch and then go from there. I was actually thinking of the entries to docs/api_doc.rst, but I think I also forgot adding some things there so I can do that later, and I'll also take a look at the sequence labeler after merging.

msperber:
And yeah, I saw that some of my commits probably broke some examples. I'll fix those after merging this.

msperber:
I took a look at the sequence labeler, and I think you are right that there was some unnecessary masking. Should be working now.

xnmt/seq_labeler.py
```diff
@@ -59,30 +62,15 @@ def _encode_src(self, src):
...
-          trg = batcher.mark_as_batch([trg_sent.get_padded_sent(token=vocab.Vocab.ES, pad_len=pad_len) for trg_sent in trg])
-          if old_mask:
-            trg.mask = np.pad(old_mask.np_arr, pad_width=((0,0), (0,pad_len)), mode="constant", constant_values=1)
+        trg = self._cut_or_pad_targets(seq_len, trg)
       else:
         raise ValueError(f"src/trg length do not match: {seq_len} != {len(trg[0])}")
+
     ref_action = np.asarray([sent.words for sent in trg]).reshape((seq_len * batch_size,))
-    loss_expr_perstep = dy.pickneglogsoftmax_batch(masked_outputs, ref_action)
+    loss_expr_perstep = dy.pickneglogsoftmax_batch(outputs, ref_action)
```
neubig: This should use `scorer.calc_loss()`, correct?

neubig:
Regarding the unit tests, I don't care strongly either way. I think any config file that requires xnmt to be run with a special setting in order to function should be documented as such though. Perhaps adding a comment to the top of the files in `test/config` that require the unit test flag would do the trick?

Feel free to merge this branch into `generalized-translator` when you're comfortable. Then I'll take a final pass through.

xnmt/seq_labeler.py
msperber: Thanks, done.

PR #442, opened 2018-06-25
[WIP] Support for multiple inputs

### LINKED ISSUE #347: Multiple source sentences
Here I go asking for things again.

It would be nice to have a way to condition on multiple input sentences and ensemble the translations with either the same model or a different model.

What would be the easiest/most straightforward way for me to add this functionality to xnmt.

I'm think of being able to specify multiple source files at inference (assuming they are aligned and have the same number of sentences), a list of models corresponding to each input file, with references if I want to use the same model, and optionally an ensembilng method (eg average logits vs average log probabilities)

msperber:
([WIP] because this depends on #429 which has not yet been merged)

This adds support for specifying multiple training src or trg files, read by multiple input readers. Should fix #347 .

To do so, I added a ```CompoundReader``` that takes in several ```InputReaders```, each of which reads one file given in a list of filenames. The result is a ```CompoundInput```.

This is mostly straight forward, but there are a few tricky places which I'm not sure if there's a better way to solve things:
- batcher does an explicit check for compound inputs, which is not a super modular solution. It's also not clear to me whether the result should be a tuple of batches or a batch of tuples: https://github.com/neulab/xnmt/blob/multi-input/xnmt/batcher.py#L350
- inference also needs to perform explicit checks if we're dealing with compound inputs or not, which is again not very modular design: https://github.com/neulab/xnmt/blob/multi-input/xnmt/inference.py#L109

Also included are a few minor changes:
- PlainTextReader supports reading the sentence length instead of its content
- getting rid of ```include_vocab_reference``` which I think was discussed sometime in the past that it would better be solved by specifying a reference to the vocab in the component that actually needs to know the vocab.


msperber:
closing in favor of #445

PR #444, opened 2018-06-26
support writing out nbest list

msperber:
Beam search already supported creating an n-best list by setting ```one_best=False```. This PR causes this n-best output to be written to file if given.


msperber:
closing in favor of #446

PR #445, opened 2018-06-27
Support for multiple inputs v2

### LINKED ISSUE #347: Multiple source sentences

msperber:
(this is a rebase of #442 )

This adds support for specifying multiple training src or trg files, read by multiple input readers. Should fix #347 .

To do so, I added a CompoundReader that takes in several InputReaders, each of which reads one file given in a list of filenames. The result is a CompoundInput.

This is mostly straight forward, but there are a few tricky places which I'm not sure if there's a better way to solve things:

batcher does an explicit check for compound inputs, which is not a super modular solution. It's also not clear to me whether the result should be a tuple of batches or a batch of tuples: https://github.com/neulab/xnmt/blob/multi-input2/xnmt/batcher.py#L350
inference also needs to perform explicit checks if we're dealing with compound inputs or not, which is again not very modular design: https://github.com/neulab/xnmt/blob/multi-input2/xnmt/inference.py#L109
Also included are a few minor changes:

PlainTextReader supports reading the sentence length instead of its content
getting rid of include_vocab_reference which I think was discussed sometime in the past that it would better be solved by specifying a reference to the vocab in the component that actually needs to know the vocab.


neubig:
Thanks! This looks good in general.

First, could you add a couple test cases demonstrating the "reads from only one file" and "reads from multiple files" behaviors, so I can get a better idea of what these would look like in a config file?

msperber:
I've added a simple unit test to demonstrate both use cases, but it's not a full experiment config because there is currently no model that actually supports multiple inputs (I've only tested it with a prototype model in my local repo). What would happen by specifying the compound reader is currently that the ```src``` argument in ```TrainableModel.calc_loss()``` and ```GeneratorModel.generate()``` will be a tuple of batches instead of a single batch.

msperber:
Update: I've addressed some of my own criticism from above by introducing a CompoundBatch which is now passed around, instead of a tuple of batches. This removes the need to check types in a few places and seems like a cleaner solution. In addition, I've introduced sent_len() and batch_size() methods to batch objects, as well as sent_len() methods to input objects. This makes working with single inputs vs batches much more convenient. It also makes it easier to read things like ```len(src)``` which should now be either ```src.sent_len()``` or ```src.batch_size()```.

philip30:
Seems OK for me in general. This is a bit difficult problem, so I agree that there will be many places need to be changed. Do you still have some issue you found yourself?

msperber:
Thanks. From my side there are no bigger issues left. The Batcher needed to be adjusted, but I feel better about it now because it will always be returning a batch, either list batch or compound batch.

PR #448, opened 2018-06-29
A few changes to support input of large files

neubig:
* Print a message when the file starts loading
* Count lines more efficiently
* Cache line counts so they don't need to be re-counted every epoch

### DIFF
--- a/xnmt/input_reader.py
+++ b/xnmt/input_reader.py
@@ -114,7 +114,7 @@ class PlainTextReader(BaseTextReader, Serializable):
 
   @serializable_init
   def __init__(self, vocab=None, include_vocab_reference=False):
-    self.super().__init__()
+    super().__init__()
     self.vocab = vocab
     self.include_vocab_reference = include_vocab_reference
     if vocab is not None:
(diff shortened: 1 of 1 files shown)


xnmt/input_reader.py
```diff
@@ -68,11 +71,14 @@ def read_sent(self, line: str) -> xnmt.input.Input:
     raise RuntimeError("Input readers must implement the read_sent function")
 
   def count_sents(self, filename):
-    f = open(filename, encoding='utf-8')
-    try:
-      return sum(1 for _ in f)
-    finally:
-      f.close()
+    if filename in self.line_counts:
```
msperber: There's a recent Python built-in annotator that does this and would save quite a few lines of code. One would simply put ```@functools.lru_cache``` before the method definition. See also 
https://docs.python.org/3.6/library/functools.html#functools.lru_cache

neubig:
Fixed, that's a really cool feature :)

msperber:
Yeah :) It might be worth adding in some other places as well, for examples methods that return DyNet expressions..

PR #449, opened 2018-06-30
Improve reporting

msperber:
This attempts to make reporting more reusable by introducing dedicated reporter classes. Currently, only an HTML attention reporter is implemented, but others could be added more easily now. I've commented out some of the related code from the segmentation models because I'm not sure how they work and to what extent they need to be maintained? In any case, it should be easy to re-add this functionality if needed.

### DIFF
--- a/examples/14_report.yaml
+++ b/examples/14_report.yaml
@@ -3,6 +3,7 @@ report: !Experiment
   exp_global: !ExpGlobal
     default_layer_dim: 256 
     dropout: 0.0
+    compute_report: True
   model: !DefaultTranslator
     src_reader: !PlainTextReader
       vocab: !Vocab {vocab_file: examples/data/head.ja.vocab}
@@ -43,5 +44,7 @@ report: !Experiment
       ref_file: examples/data/head.en
       hyp_file: examples/output/{EXP}.test_hyp
       inference: !AutoRegressiveInference
-        report_path: examples/output/{EXP}.report
-        report_type: html, file
+        reporter:
+        - !AttentionHtmlReporter { report_path: 'examples/output/{EXP}.report' }
+        - !AttentionFileReporter { report_path: 'examples/output/{EXP}.report' }
+
--- a/examples/14_report.yaml
+++ b/examples/14_report.yaml
@@ -1,42 +1,26 @@
-# This example demonstrates reporting of attention matrices.
+# XNMT supports writing out reports, such as attention matrices generated during inference.
+# These are generally created by setting exp_global.compute_report to True, and adding one or several reporters
+# to the inference class.
 report: !Experiment
   exp_global: !ExpGlobal
-    default_layer_dim: 256 
-    dropout: 0.0
     compute_report: True
   model: !DefaultTranslator
     src_reader: !PlainTextReader
       vocab: !Vocab {vocab_file: examples/data/head.ja.vocab}
     trg_reader: !PlainTextReader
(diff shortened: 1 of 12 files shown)


philip30:
I think this design was once avoided by @neubig as we want to let each component generate the report by themselves. Did you also handle the reporting in segmenting encoder correctly?

msperber:
I see.. could you explain the reason for this a little further? Using this PR's design, it's still easy to write specialized reporting code by simply introducing a specialized class.

To explain my motivation for this PR: I'm currently working with an attentional model that is not based on the default translator, and when I wanted to use reporting I ran into some difficulties:
1. It was hard to understand the previous reporting code, partly because information was passed around from different places, and partly because the ```Reportable``` base class contained a lot of undocumented code.
2. I didn't want to copy-and-paste the reporting code as attention reporting seems like a generally useful feature (e.g. also useable with self-attention layers, the fixed-size attention layer, etc.), and duplicating code is usually not a good idea and makes things hard to maintain.

The segmenting reporting is currently commented out; I could certainly draft a specialized reporter for it according to the proposed interface.

philip30:
Yea, I think it is dated on 300 PRs ago at #147 when we firstly introduce the Hierarchical Model class which was refactored to xnmt.events. At that time the interface changes constantly, so sorry for the lack of documentation!

The reason for that was:
1. We don't want to copy paste code to handle reporting (as what you said) because we change a single component in XNMT. Generally the reporting scenario should also change automatically when we change some component.

For example the segmenting encoder can generate some segmentation biproduct both the decoder can produce the attention matrix. In your proposed design, so you would basically inherit the HTMLReporter and handle the segmentation, right? But it means that we always need to know which kind of reporter compatible to which kind of component we use in XNMT. I think this will be harder to maintain?

2. Basically if you want to generate a report (particularly HTML), you just need to build the html context. If it is the root, the context should be None, so it should return a new HTML root. But if it is the children (for example, the encoder was once the descendant of Translator, because Translator has a reference to encoder), the context will not be None, so we should need to modify the context.

msperber:
Thanks for the explanation, and no problem about the documentation!

I've pushed some draft code to show how reporting segmentations would look like. It's untested because there doesn't seem to be a unit test for the SegmentingSeqTransducer, but the basic idea should be clear. We could certainly introduce a base class for HTML reporting that handles some of the boilerplate, but I would leave that for a later PR. The point of this PR is mainly to establish a consistent and reusable reporting interface. It's true that we have to know which reporter goes with which model, but I think that could be easily documented in the reporter's doc string, and the same also holds for many other parts of XNMT such as matching embedder / input reader, matching generator model / inference class, etc.

As for HTML files, my code would currently result in separate HTML files for each reporter, although it shouldn't be too hard to refactor this into a reporter that puts everything in the same HTML file if this is really needed (I would also leave this for a future PR though).

philip30:
Can the new code generate report for both translator and the segmenting
encoder? If not, I think there is still some functionality missing from the
previous code. Yes, I have been writing the unitttest for my code but it is
still unmerged. Iwill have more time after my thesis presentation though..

msperber:
Yes it can, be simply specifying:
```
  reporter:
  - !AttentionHtmlReporter {}
  - !SegmentingHtmlReporter {}
```

And I think it's fine to add tests later when you have time. I've already put comments in the code that the segmenting reporter is still untested.

philip30:
It looks much cleaner I think (than the previous). LGTM if this design is OK for @neubig

Issue #436, opened 2018-06-22
Language model training objective


neubig:
It would be nice to have the ability to train a language model within xnmt, for things like this:

http://aclweb.org/anthology/D/D17/D17-1039.pdf

neubig:
Maybe can be assigned to @msperber ?

msperber:
Yes, I'm planning to take a look at this.

neubig:
Should be done by #452

PR #452, opened 2018-07-04
Language Model

msperber:
This introduces a language model that is a subclass of ```TrainableModel``` (not ```GeneratorModel```). It can be trained and used with the loss evaluator (but not the accuracy or decoding evaluators).

The code is straight forward, the main question to discuss is handling of source / target inputs. The XNMT training code is currently built around the assumption that we always have a source side and a target side. The language model requires only a source side. I can think of a few ways to handle this:

- Current solution: simply configuring the trg_file identical to the src_file, and then ignoring the trg sentences in the model. A bit hacky, but requires no changes to the training and evaluation code.
- Make a non-parallel version of ```SimpleTrainingTask``` that reads only source inputs and does not use ```read_parallel_corpus```. In some sense the cleanest solution, but would introduce quite a bit of new code.
- Make trg_file / trg_reader optional, and modify ```read_parallel_corpus``` to return some kind of dummy target sentences or something similar. This would fall somewhere between the above 2 solutions.

Addresses #436

### DIFF
--- a/test/config/lm.yaml
+++ b/test/config/lm.yaml
@@ -1,0 +1,27 @@
+lm: !Experiment
+  model: !LanguageModel
+    src_reader: !PlainTextReader
+      vocab: !Vocab {vocab_file: examples/data/head.en.vocab}
+    src_embedder: !SimpleWordEmbedder
+      emb_dim: 512
+    rnn: !UniLSTMSeqTransducer
+      layers: 1
+    scorer: !Softmax
+      vocab: !Vocab {vocab_file: examples/data/head.en.vocab}
+  train: !SimpleTrainingRegimen
+    batcher: !SrcBatcher
+      batch_size: 32
+    trainer: !AdamTrainer
+      alpha: 0.001
+    run_for_epochs: 2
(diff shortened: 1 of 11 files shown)


msperber:
If we decide for (2), a possible class hierarchy would be:

- TrainingTask
  - SupervisedTrainingTask
  - UnsupervisedTrainingTask
- TrainableModel
  - UnsupervisedModel
  - SupervisedModel
    - GeneratorModel

SimpleTrainingRegimen could either be changed to accept a (supervised or unsupervised) training task subobject, similar to how multi task regimens take several tasks objects. Or, it could be split into SimpleSupervisedTrainingRegimen and SimpleUnsupervisedTrainingRegimen.

For comparison, the current hierarchy is:
- TrainingTask
  - SimpleTrainingTask
- TrainableModel
  - GeneratorModel
- TrainingRegimen
  - SimpleTrainingRegimen
  - several multi task regimens

msperber:
Thanks for the review. I'll merge this then and leave the further extensions / clean-up for future PRs.

PR #453, opened 2018-07-06
fix sentence piece input reader byte string

cindyxinyiwang:
Just add one line to convert sentence piece encoded bytes into string


msperber:
LGTM, I'll merge this

neubig:
One thing: is this dependent on the version of sentencepiece? Will it break things if an old version is installed?

PR #454, opened 2018-07-06
Better computation time tracking for same-batch multi tasking

msperber:
Previously, the same-batch regimen computed time/words jointly for all tasks. This PR computes and logs times individually if per_task_backward is activated.

In addition, includes a minor cleanup by removing the update_weights flag from the training regimens which was only used in one unit test, where the same effect is now achieved via a new DummyTrainer.


msperber:
I'll merge this also as this is a straight-forward fix / cleanup.

PR #455, opened 2018-07-06
print current path when an exception in a constructor occurs

msperber:


msperber:
I don't think this needs a review so I'll merge this

PR #456, opened 2018-07-06
fix case of no input feeding

msperber:


msperber:
This also won't need a review.

PR #462, opened 2018-07-10
Cascaded inference + updated model interface

msperber:
This implements a cascaded generator and inference object that can be used to chain two or several models. A cascade is a generator model but not trainable, so I slightly updated to model class hierarchy as follows:
- TrainableModel
  - UnsupervisedModel
  - SupervisedModel
- GeneratorModel

This hierarchy is also the first step toward better support for unsupervised models such as language models (#452 ). Once this and the language model are merged into master, I would take care of the distinction between supervised / unsupervised training tasks.

Using the cascade in practice would look as follows:
```
  ...
  evaluate:
    - !AccuracyEvalTask
      eval_metrics: bleu
      src_file: examples/data/head.ja
      ref_file: examples/data/head.en
      hyp_file: examples/output/{EXP}.test_hyp
      model: !CascadeGenerator
        generators:
        - !Ref { path: model }
        - !Ref { path: model }
      inference: !CascadeInference
        steps:
        - !AutoRegressiveInference {}
        - !AutoRegressiveInference {}
```

### DIFF
--- a/xnmt/model_base.py
+++ b/xnmt/model_base.py
@@ -28,7 +28,7 @@ class TrainableModel(object):
     """
     raise NotImplementedError("Pick a key for primary loss that is used for dev_loss calculation")
 
-class UnsupervisedModel(object):
+class UnsupervisedModel(TrainableModel):
   """
   A template class for an unsupervised trainable model, implementing a loss function based on inputs only.
 
@@ -54,7 +54,7 @@ class UnsupervisedModel(object):
     """
 
 
-class SupervisedModel(object):
+class SupervisedModel(TrainableModel):
   """
   A template class for a supervised trainable model, implementing a loss function based on inputs and outputs.
 
--- a/xnmt/model_base.py
+++ b/xnmt/model_base.py
@@ -1,6 +1,6 @@
 from typing import Sequence, Union
 
-from xnmt import batcher, events, input, input_reader, loss, output, training_task
+from xnmt import batcher, events, input_reader, loss, output, training_task
 import xnmt.loss_calculator
 import xnmt.input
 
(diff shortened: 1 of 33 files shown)


xnmt/model_base.py
```diff
@@ -1,12 +1,62 @@
...
+  def get_primary_loss(self) -> str:
+    """
+    Returns:
+      Identifier for primary loss.
+    """
+    raise NotImplementedError("Pick a key for primary loss that is used for dev_loss calculation")
+
+class UnsupervisedModel(TrainableModel):
+  """
+  A template class for an unsupervised trainable model, implementing a loss function based on inputs only.
```
neubig: I think "Unsupervised" is a bit of a misnomer here. I'd prefer something like "Monolingual" (although I'm not completely satisfied with this either). Also, "based on inputs only" seems a little strange. Maybe "based on a single input" or something?

xnmt/model_base.py
```diff
@@ -1,12 +1,62 @@
...
+      loss_calculator: loss calculator.
+
+    Returns:
+      A (possibly batched) expression representing the loss.
+    """
+
+
+class SupervisedModel(TrainableModel):
+  """
+  A template class for a supervised trainable model, implementing a loss function based on inputs and outputs.
```
neubig: Similarly, what about "Bilingual"?

msperber:
I can see the point.. Given that I'd also add the corresponding training tasks in a later PR, would ConditionedTrainingTask / UnconditionedTrainingTask work for you as well?

neubig:
Sure.

msperber:
Done, I'll merge.

PR #464, opened 2018-07-11
Several miscellaneous robustness fixes

neubig:
This PR has a bunch of fixes for robustness when doing lots of experiments. Probably only one is debatable, commenting out the initial model saving when starting training (`experiment.py`). I actually wasn't sure why this was necessary in the first place, and it can significantly slow down the start of training (and by proxy, the speed of debugging implementation).

### DIFF
--- a/xnmt/experiment.py
+++ b/xnmt/experiment.py
@@ -51,7 +51,7 @@ class Experiment(Serializable):
     eval_scores = ["Not evaluated"]
     if self.train:
       logger.info("> Training")
-      save_fct() # save initial model
+      # save_fct() # save initial model
       self.train.run_training(save_fct = save_fct)
       logger.info('reverting learned weights to best checkpoint..')
       ParamManager.param_col.revert_to_best_model()
--- a/xnmt/experiment.py
+++ b/xnmt/experiment.py
@@ -4,7 +4,7 @@ from xnmt import logger
 from xnmt.exp_global import ExpGlobal
 from xnmt.eval_task import EvalTask
 from xnmt.model_base import GeneratorModel
-from xnmt.param_collection import ParamManager
+from xnmt.param_collection import ParamManager, RevertingUnsavedModelException
 from xnmt.preproc_runner import PreprocRunner
 from xnmt.training_regimen import TrainingRegimen
 from xnmt.persistence import serializable_init, Serializable, bare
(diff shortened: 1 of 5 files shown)


msperber:
I’ll take a look later, but the initial saving of the model was mainly because in some cases, particular when testing or debugging, I ran into situations where the program wanted to revert back to the best model without ever having saved it (eg because the training loop was skipped). I’m not completely sure if this situation can still occur though.

xnmt/optimizer.py
```diff
@@ -34,10 +34,16 @@ def update(self) -> None:
...
+    try:
+      if not (self.skip_noisy and self._check_gradients_noisy()):
+        self.optimizer.update()
+      else:
+        logger.info("skipping noisy update")
+    except RuntimeError:
+      logger.warning("Failed to perform update. Skipping example and clearing gradients.")
+      for subcol in ParamManager.param_col.subcols.values():
+        for param in subcol.parameters_list():
+          param.scale_gradient(0)
```
msperber: This looks interesting.. will this be able to skip updates that would have lead to NaNs?

xnmt/experiment.py
```diff
@@ -51,7 +51,7 @@ def __call__(self, save_fct):
     eval_scores = ["Not evaluated"]
     if self.train:
       logger.info("> Training")
-      save_fct() # save initial model
+      # save_fct() # save initial model
```
msperber: Does this work even when setting ```run_for_epochs``` to 0? If so I think it's ok to out-comment this for now. Supporting ```run_for_epochs=0``` is important for inference-only configs because with multi-tasking one can't set ```train=None``` because the models are sub-objects of the training regimen.

philip30: maybe we just save the model in the above cases? I find it is also annoying to wait at the beginning of training...

msperber: Yeah you're right.. actually I think there's no need to save an inference-only model in the first place. I changed the code to skip the ```revert_to_best_model()``` in case it has never been saved.

xnmt/optimizer.py
neubig: Yes in theory. Unfortunately I haven't been able to test it though, as the bug causing NaNs seems to have resolved itself.

neubig:
OK, I think this should be good, right? I'll go ahead and merge.

PR #465, opened 2018-07-15
Modularize transformer encoder

neubig:
This is a first step towards a flexible version of the Transformer that is not a monolithic model, but rather model created from a sum of its parts. This is on the encoder side only at the moment. The decoder side requires more changes that will touch other parts of xnmt, so I'd like to do it in a different PR. See the specification of a 2-layer transformer model (hopefully faithful?) here: https://github.com/neulab/xnmt/blob/modularize-transformer/examples/21_self_attention.yaml#L33

One thing you'll note is that it's a bit verbose. There are a couple of reasons:
* I'm creating an MLP from two (NonLinear and Linear) transforms, which could potentially be a single `SeqTransformer`.
* Each layer has to be specified repeatedly. This will be annoying when we have a 6-layer model.

One question for Matthias: do you think it would be possible to have a `RepeatedSeqTransducer` or something like this? It would basically be a modular sequence transducer that repeats the same specification over and over again. This seems trivial in the current setup if the layers share parameters (you'd just write a for loop and call `transduce()` multiple times), but it's non-trivial when the models do not share parameters, as you'd have to call the same constructor multiple times. I wonder if there's a way to specify this in the serializer?


neubig:
Cool! I think the `!Repeat` syntax would be perfect, would you mind drafting it @msperber ?

Issue #467, opened 2018-07-16
!Repeat syntax


msperber:
A ```!Repeat``` syntax would allow specifying multilayer architectures where the same layer spec is repeated several times in a less annoying way. The idea would be to resolve this in the preloading step. A draft syntax would be:
```
!Repeat
  times: 6
  shared: False
  content: !ModularSeqTransducer
    modules:
    - ResidualSeqTransducer { ... }
    - ResidualSeqTransducer { ... }
```

PR #468, opened 2018-07-16
initial implementation of repeat syntax

msperber:
As mentioned in #467.

A test of a simple configuration with the following encoder seemed to work:
```
    encoder: !ModularSeqTransducer
      modules: !Repeat
        times: 3
        content: !BiLSTMSeqTransducer {}
```

@neubig could you take a look if this will work as expected for the multilayer transformer configs?


neubig:
Great, it was this simple :)

PR #469, opened 2018-07-17
Added MLP convenience class

neubig:
This makes the transformer encoder more compact by using the repeat syntax and adding an MLP Transform convenience class that performs 0 or more non-linear transforms, then a linear transform.


philip30:
I think it's better to just use 'MLP' name? Most of the users can tell by just the name I think.

neubig:
Yeah, that's probably right. I was originally a little bit averse to having a class name that's just an abbreviation, but it's certainly easier to write. I've fixed it.

PR #470, opened 2018-07-17
Updated coding conventions

msperber:
This updates coding conventions by (1) improving structure a bit, and (2) making some things that we were already following more explicit, regarding doc strings, type annotations, and unit tests.


neubig:
Nice! Thanks.

PR #474, opened 2018-07-18
Reporting features

msperber:
This PR makes the reporting feature more powerful:
- add a ReferenceDiffReporter that visualizes diffs between system output and reference using character-sequence matching. This is based on the charcut tool from https://github.com/alardill/CharCut
- add a CompareMtReporter that simply calls the compare-mt.py script from https://github.com/neubig/util-scripts
- HtmlReporter / AttentionReporter are extended:
  - better support for UTF-8 (correctly displayed in HTML; display romanized strings in attention plots)
  - integrate reference diff visualization for each sentence
  - various cosmetic improvements
- better file structure: root dir is ```reports/{EXP_NAME}/``` and contains only the main files; images etc. are in subdirs.
- added a new ```xnmt.thirdparty``` package which contains some of the added code, and I also moved the ```speech_features``` code there


pierregodard:
@msperber: Can you confirm that plain text reporting capabilities for attention matrices disappeared in recent refactorings? Should I create an `AttentionTxtReporter` class, inheriting from a new `TxtReporter` class in order to be consistent with what's done for HTML? This might be overkill but better than a dirty hack on the HTML classes, right?

msperber:
You are right @pierregodard, it seems the text reporting got lost. Yes, I also think the proper way to re-add it would be by adding a new ```AttentionTxtReporter```.

pierregodard:
Thank you @msperber, I'll proceed to do that and make a PR soon then.

msperber:
I'll merge this because I would like to work on related issues without creating merge conflicts, and because this is mainly adding new things and unlikely to break anything. I'll be happy to reflect any comments in a follow-up commit.

PR #475, opened 2018-07-19
Simplify vocab

msperber:
#267 introduced the convention that vocabs are fixed from initialization time in order to ensure consistency with multi task training, handling of multiple training corpora, etc.

This PR simplifies the vocab by removing some of the older code that was needed for dynamically growing vocabs but is no longer needed. Besides simplifying the code, this solves consistency problems when freeze() would not be called or called too late in cases where the vocab is not used immediately at the start of the training. It also updates some tests that hadn't followed the convention correctly.


neubig:
Looks good!

PR #480, opened 2018-07-25
log unhandled exceptions to file

### LINKED ISSUE #458: Errors on crashes should be written to log files
Currently when xnmt crashes, the exception is not written to the log file. This makes it difficult to track down and should be fixed.

msperber:
fixes #458

### DIFF
--- a/xnmt/__init__.py
+++ b/xnmt/__init__.py
@@ -12,6 +12,7 @@ if package_dir not in sys.path:
 import logging
 logger = logging.getLogger('xnmt')
 yaml_logger = logging.getLogger('yaml')
+file_logger = logging.getLogger('xnmt_file')
 
 import _dynet
 dyparams = _dynet.DynetParams()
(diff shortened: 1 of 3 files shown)


msperber:
Update: seems that this is still not fool-proof. It does write more to the log file than before though..

xnmt/__init__.py
```diff
@@ -12,6 +12,7 @@
 import logging
 logger = logging.getLogger('xnmt')
 yaml_logger = logging.getLogger('yaml')
+file_logger = logging.getLogger('xnmt_file')
```
neubig: I didn't understand why you need another logger for this (does this imply a separate log file for crash logs?). Can't you just use the existing one? I think the best thing would be to have the logging information at the end of the standard log file. It's possible that this is what already happens and I just didn't understand the code though.

msperber: Yes that's what's happening already. I just needed to catch unhandled exceptions and log them explicitly, and using the normal logger this would have resulted in the exception appearing on the console output twice, so I created another logger that logs only to file (the same file as the standard logger), not to the console.

neubig: OK, cool. Happy to merge this then.

PR #483, opened 2018-07-26
Batch norm

msperber:
This PR adds batch norm. I've been using it in my private branch for a long time and think there should be no issues. I'll send some model code that actually uses batch norm later.


msperber:
Sure, the most convenient way to do this was to also add the network-in-network transducer which I had planned for a follow-up PR but added to this one now. I added a test under test/config/encoders.yaml

msperber:
1. Hm yes we could modularize this further, although we would have to think about the proper interface a bit: Would we have something like a ```LinearSeqTransducer``` followed by ```BatchNorm``` and a ```NonLinearitySeqTransducer```? The first and last seem related to classes from the ```transforms``` module. Maybe we can introduce a helper class that turns any transducer (object that accepts and returns DyNet expressions) into a sequence transducer? One complication is masking, though. For the Nin layer, it's not necessary to mask the linear projection, only the output after the nonlinearity needs to be masked. Also, batch norm would need to be made a sequence transducer, but that should be easy. What do you think?
2. What would be the advantage of doing this? The layer norm class is only a few lines and I think it would become much harder to read by merging both classes (or maybe I misunderstood?)..
3. It's a bit slow but not unusable, my bigger concern is that it's wasting quite a bit of memory. Having some dedicated DyNet operations would be very helpful. One complicating factor is the proper masking which is quite important. The submitted code has been working for me though, so I wouldn't plan on working on this myself.

neubig:
1. We already have this for the transforms `TransformSeqTransducer`. Batch Norm would have to be implemented.
2. Sorry, that's not what I meant. I just meant that we'd create a `SeqTransducer` that would perform layer norm or batch norm. But now that I think about it, would it be OK to have layer norm and batch norm be `Transform` classes? If so, we could just use `TransformSeqTransducer` for all of them.
3. OK, sounds good.

msperber:
Forgot to mention: I also extended ```TransformSeqTransducer``` to pass on the mask, and to support downsampling (via reshapes).

xnmt/transducers/network_in_network.py
```diff
@@ -0,0 +1,47 @@
+from xnmt.transducers import base
+from xnmt.modelparts import transforms as modelparts_transforms
+from xnmt.persistence import Serializable, serializable_init, bare, Ref
+from xnmt import norms, param_initializers
+
+class NinLayer(base.ModularSeqTransducer, Serializable):
```
neubig: Maybe `NinSeqTransducer`?

Issue #439, opened 2018-06-25
Better import conventions to avoid circular imports


msperber:
I have had problems with circular imports a number of times which can be annoying. More exhaustive type annotations (#438 ) might make this a more prominent issue.

It seems that most such issues can be avoided by importing modules instead of types from inside a module. E.g. instead of writing ```from xnmt.input import SimpleSentenceInput```, one would write ```import xnmt.input``` or something similar. This is already done in some places, but not consistently. See also https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python

There are 2 things to keep in mind:
- Doing so makes things more verbose. We probably want to use ```from xnmt import input``` / ```a=input.SimpleSentenceInput()``` instead of the more verbose ```import xnmt.import``` / ```a=xnmt.import.SimpleSentenceInput()```.
- A lot of the modules have unfortunate names. E.g. ```input``` is a reserved Python keyword, the ```batcher``` and ```decoder``` modules interfere with variables / arguments of the same name, etc. I would propose to rename the modules to more general names, e.g. ```batching```, ```inputs```, ```decoding```, etc.

msperber:
Are there any comments / objections to the above suggestions? I'm getting an increasing number of import issues and would like to address this soon, though probably after #380 is merged to avoid merge conflicts there.

philip30:
Hey sorry for the delay. I will be working on that soon.

neubig:
I'm a little bit worried about the verbosity, but if there's no better way then I'm fine with this.

msperber:
resolved by #550.

Issue #438, opened 2018-06-25
Exhaustive type annotation


neubig:
One of the problems I currently perceive with xnmt is that it's not clear what the inputs and outputs of each function are supposed to be. This is important because xnmt is a relatively big project written in a modular, object-oriented way that should enforce this. I think we should aim for exhaustive type annotation if possible to make this more clear.

One way to do so would be using mypy, and PyAnnotate:
> http://mypy-lang.org
> https://github.com/dropbox/pyannotate

I've tried mypy out and it looks good, although it's not clear whether it plays well with other external libraries.

We should probably also add type annotation to the programming style guide:
http://xnmt.readthedocs.io/en/latest/programming_style.html

msperber:
Yeah, I think using those tools would be a good idea. This would basically be a one-time undertaking, right?

One potential difficulty we might face is circular imports, because type annotations requires importing the respective types. I have some ideas about this which I'll put in a separate issue.

neubig:
Yes, this would be a one-time thing. And thanks!

PR #484, opened 2018-07-27
Improved import conventions Pt 1

msperber:
Better import conventions will be helpful for avoiding circular imports, which are already a problem and will become even more so when adding exhaustive type annotations (#438).

As suggested in #439, this PR starts by renaming modules to avoid name clashes with variable names. For example, ```batcher``` is a well-used variable name, which prevents one from doing ```from xnmt import batcher``` because the module name will be shadowed by the variable name. This is probably the less debatable part of #439 because it reduces verbosity in most places rather than adding to it. E.g. ```import xnmt.batcher; xnmt.batcher.mark_as_batch()``` can now be replaced by ```from xnmt import batching; batching.mark_as_batch()```.

When renaming modules, I tried to be somehow consistent, but the main goal was to avoid name clashes with common variable names while keeping module names short. Let me know if you find anything confusing or have better naming suggestions.


philip30:
Hmmm, it's hard to guess the correct module name in the future. There is abbreviated module (actually a file, python module is a whole folder), but there is also truncated name, or full name... I think the cleaner solution is using the full module path, but just rename them at the start of function/files, and we can guarantee to avoid clash. When the code get more bigger, it's hard to find not colliding name without using an escape like "_" for local variable. It's just my opinion.

msperber:
Hmm it's true that we can't avoid all name clashes, but I think we should at least try as it is currently happening very frequently. But I can see your point that the names should make it easy to find classes in the code base... I made another attempt by using the main classes name as module name but then making it plural where appropriate, e.g. ```attenders``` and ```batchers```. This would be a smaller change to the names than before and should be easy to understand and avoid most name clashes..

msperber:
Sure, I'll finalize this very soon. ```long_short_term_memory.py``` seems unnecessarily long, so I'd just call it ```recurrent.py```.

neubig:
What about `recurrent_transducers` or something? I think it's OK to call it just `recurrent` as well (that's certainly less verbose), but as @philip30 pointed out, having some things plural some things not will make remembering the module names difficult.

msperber:
Ok, done.

neubig:
Thanks, merge if you're ready.

PR #486, opened 2018-07-30
Started refactoring loss calculation [WIP]

neubig:
I started taking a look at implementing the decoder side of the Transformer, and realized that there is a bunch of functionality in `DefaultTranslator` that makes it a bit convoluted and difficult to implement there. I started refactoring to try to make DefaultTranslator simpler while not removing too much default functionality. I'm still not finished, but wanted to post here in case @philip30 or @msperber has feedback. Here is a summary of the changes:

* Previously, the calculation of the loss for each translator was quite complicated for MLE. Basically it called `model.calc_loss(src, trg, loss_calculator)`, which then called `loss_calculator.calc_loss(model, src, trg)`, which looped through calls to `model.calc_loss_one_step()`. I changed this to just `loss_calculator.calc_loss(model, src, trg)`. This is nice because it allows for code separation where it's possible. 
* I removed ability to handle `CompoundExpressionSeq`s in the translator, which makes things significantly more difficult. Instead, `MLELoss` can repeat identical calculation multiple times. If there is non-determinism in the encoder, for example, this will have the same affect. This greatly simplifies things, but also has the effect that some calculations might be unnecessarily performed multiple times. I'd suggest caching heavy computations within each of the components, then if the component gets an input that it's seen before, it will just re-use the output.
* I tried to remove calls that were accessing arbitrary data members of the translator.

There are several things left to do:
- [ ] Change Reinforce and MinRisk losses over to the new method of calculation (easy?)
- [ ] The global fertility function has been removed from the loss, and probably needs to be re-added.
- [ ] Resolve all unit tests and make sure that things are working.

Either way, first if you could take a look and see if this makes sense @philip30, that'd be nice!

### DIFF
--- a/test/config/seg_report.yaml
+++ b/test/config/seg_report.yaml
@@ -13,11 +13,12 @@ report: !Experiment
       final_transducer: !BiLSTMSeqTransducer {}
       policy_learning: !PolicyGradient 
         z_normalization: True
-        sample: 2
   train: !SimpleTrainingRegimen
     run_for_epochs: 1
     src_file: examples/data/head.ja
     trg_file: examples/data/head.en
+    loss_calculator: !MLELoss
+      repeat: 2
   evaluate:
     - !AccuracyEvalTask
       eval_metrics: bleu,wer
--- a/test/test_decoding.py
+++ b/test/test_decoding.py
@@ -10,7 +10,6 @@ from xnmt.decoder import AutoRegressiveDecoder
 from xnmt.embedder import SimpleWordEmbedder
 import xnmt.events
 from xnmt.input_reader import PlainTextReader
-from xnmt.loss_calculator import AutoRegressiveMLELoss
 from xnmt.lstm import UniLSTMSeqTransducer, BiLSTMSeqTransducer
 from xnmt.param_collection import ParamManager
 from xnmt.transform import NonLinear
@@ -87,9 +86,8 @@ class TestForcedDecodingLoss(unittest.TestCase):
 
   def test_single(self):
     dy.renew_cg()
-    train_loss = self.model.calc_loss(src=self.src_data[0],
-                                      trg=self.trg_data[0],
-                                      loss_calculator=AutoRegressiveMLELoss()).value()
+    train_loss = self.model.calc_nll(src=self.src_data[0],
+                                     trg=self.trg_data[0]).value()
     dy.renew_cg()
     outputs = self.model.generate(xnmt.batcher.mark_as_batch([self.src_data[0]]), [0], GreedySearch(),
                                   forced_trg_ids=xnmt.batcher.mark_as_batch([self.trg_data[0]]))
@@ -128,9 +126,8 @@ class TestFreeDecodingLoss(unittest.TestCase):
     output_score = outputs[0].score
 
     dy.renew_cg()
-    train_loss = self.model.calc_loss(src=self.src_data[0],
-                                      trg=outputs[0],
-                                      loss_calculator=AutoRegressiveMLELoss()).value()
+    train_loss = self.model.calc_nll(src=self.src_data[0],
+                                     trg=outputs[0]).value()
 
     self.assertAlmostEqual(-output_score, train_loss, places=5)
 
--- a/xnmt/lm.py
+++ b/xnmt/lm.py
@@ -38,7 +38,7 @@ class LanguageModel(model_base.ConditionedModel, model_base.EventTrigger, Serial
   def get_primary_loss(self):
     return "mle"
 
-  def calc_loss(self, src, trg, loss_calculator):
+  def calc_nll(self, src, trg):
     if not batcher.is_batched(src):
       src = batcher.ListBatch([src])
 
@@ -60,7 +60,4 @@ class LanguageModel(model_base.ConditionedModel, model_base.EventTrigger, Serial
       loss_expr_perstep = dy.cmult(loss_expr_perstep, dy.inputTensor(1.0-src_targets.mask.np_arr.T, batched=True))
     loss_expr = dy.sum_elems(loss_expr_perstep)
 
-    model_loss = loss.FactoredLossExpr()
-    model_loss.add_loss("mle", loss_expr)
-
-    return model_loss
+    return loss_expr
--- a/xnmt/loss_calculator.py
+++ b/xnmt/loss_calculator.py
@@ -10,12 +10,13 @@ from xnmt.constants import INFINITY
 from xnmt.transform import Linear
 import xnmt.evaluator
 import xnmt.batcher
+import xnmt.model_base
 
 class LossCalculator(object):
   """
   A template class implementing the training strategy and corresponding loss calculation.
   """
-  def calc_loss(self, translator, initial_state, src, trg):
+  def calc_loss(self, translator, src, trg):
     raise NotImplementedError()
 
   def remove_eos(self, sequence, eos_sym=Vocab.ES):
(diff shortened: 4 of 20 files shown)


test/config/seg_report.yaml
```diff
@@ -13,11 +13,12 @@ report: !Experiment
       final_transducer: !BiLSTMSeqTransducer {}
       policy_learning: !PolicyGradient 
         z_normalization: True
-        sample: 2
   train: !SimpleTrainingRegimen
     run_for_epochs: 1
     src_file: examples/data/head.ja
     trg_file: examples/data/head.en
+    loss_calculator: !MLELoss
+      repeat: 2
```
philip30: Maybe we can set the default value of the repeat to be exp_global.enc_sample_num? This flag will bind together the number of sample in the encoder and in repeat. So we don't need to specify repeat manually.

test/test_decoding.py
```diff
@@ -10,7 +10,6 @@
 from xnmt.embedder import SimpleWordEmbedder
 import xnmt.events
 from xnmt.input_reader import PlainTextReader
-from xnmt.loss_calculator import AutoRegressiveMLELoss
```
philip30: @msperber  once renamed MLELoss to AutoRegressiveMLELoss. Maybe he has a comment of the "renaming" this back?

xnmt/lm.py
```diff
@@ -38,7 +38,7 @@ def shared_params(self):
   def get_primary_loss(self):
     return "mle"
 
-  def calc_loss(self, src, trg, loss_calculator):
+  def calc_nll(self, src, trg):
```
philip30: So MLE always needs a model that calculate nll with two arguments. This is OK so far assuming the supervised model. I guess we will never use MLE for unsupervised model, right?

xnmt/loss_calculator.py
```diff
@@ -27,66 +28,27 @@ def remove_eos(self, sequence, eos_sym=Vocab.ES):
...
-          if mask is None or mask.np_arr[j, index] == 0 or np.sum(mask.np_arr[:, index]) == mask.np_arr.shape[0]:
-            assert not found_masked, "sentences must be sorted by decreasing target length"
-            ret.append(single_trg[index])
-          else:
-            found_masked = True
-        return xnmt.batcher.mark_as_batch(ret)
-    else:
-      if not xnmt.batcher.is_batched(sent): return sent[index]
-      else: return xnmt.batcher.mark_as_batch([single_trg[index] for single_trg in sent])
+    losses = [model.calc_nll(src, trg) for _ in range(self.repeat)]
```
philip30: I see what you did here. So we loop the sample on the loss calculation and the model that do the sampling it needs to store the appropriate states for itself. I think this is a good idea and generally a cleaner solution. However, the loss should be calculated per sample, not the sum of it.

philip30: So we can deal and calculate the loss of the sample appropriately. The esum losses will delete the reference of the single loss calculation. That's why I cached the losses in this part and access that before.

xnmt/specialized_encoders/segmenting_encoder/segmenting_encoder.py
```diff
@@ -102,7 +102,14 @@ def transduce(self, embed_sent: ExpressionSequence) -> List[ExpressionSequence]:
         sent_context = self.final_transducer.transduce(expr_seq)
         self.final_states.append(self.final_transducer.get_final_states())
         enc_outputs.append(sent_context)
-      return CompoundSeqExpression(enc_outputs)
+      # TODO (neubig): This is a temporary minimial fix to revert to using only one sample.
+      #                Multiple samples can be done by the MLELoss(repeat=samp_num) functionality.
```
philip30: I can fix this if you have decided about the losses from multiple samples. So it means that nothing should really be changed in the calc_additional_loss part because it still needs a single loss from each of the sample. And the same as in policy_gradient.py

msperber:
My one comment is regarding the non-autoregressive models (currently sequence labeler and sequence classifier) which don't have loss calculators. This is currently already a bit inconsistent because they still are passed the standard loss calculator in ```calc_loss()```, but just don't use it. However, just calling ```loss_calcular.calc_loss()``` instead of ```model.calc_loss()``` in the training task is probably not general enough right now?

test/config/seg_report.yaml
neubig: Or maybe we can get rid of `enc_global.enc_sample_num`?

test/test_decoding.py
neubig: I think that's because there was a bunch of extra stuff in there that only worked for auto-regressive models. Now `MLELoss` is just a wrapper around the `calc_nll` function, and is more general.

xnmt/lm.py
neubig: I think we discussed this before when @msperber implemented the LM, and we decided that for now we'd just stick with this design (a simple work-around is to specify the same file for `src` and `trg`) but might revise it later. This is not new in this PR.

xnmt/loss_calculator.py
neubig: > "the loss should be calculated per sample, not the sum of it"
Could you elaborate? I think if there's a use case we can handle it, but we need to make sure that we're consistent about the expected input and output of each of the functions. Currently, `calc_loss` is expected to return a single loss value in all cases.

xnmt/specialized_encoders/segmenting_encoder/segmenting_encoder.py
neubig: Yeah, it'd be great if you could fix this later when I'm done with my changes to the interface. I just wanted to check with you first to make sure I didn't do anything that would make it difficult for you to do what you want to do.

neubig:
@msperber Actually, I think the design now is cleaner than before and largely fixes this problem. Basically `MLELoss` just is a call to `calc_nll()` which should work for all models. Other loss calculators will work with some subset of the models (we can add a check to make sure that they are passed a model that they are compatible with).

test/config/seg_report.yaml
philip30: I see. I think we can just stack the sample on the fly while looping. OK, I think if we can just pass the unaccumulated loss generated from the repeat, everything should be settled.

msperber:
I see, that sounds good then!

xnmt/loss_calculator.py
philip30: I mean each sample src will produce different trg. And that different trg has a different MLE loss against the reference. I need to have the separate losses (Not the sum) so I can bptt through each sample according to their loss, not the sum of the sample losses. For example in Reinforce, we need the separate evaluation for each of the sample right, not the sum of the losses.

neubig: OK, got it. Let me see what I can do with this. I think perhaps if we move the calls to `calc_additional_loss()` from `training_task.py` to `loss_calculator.py`, we should be able to solve the problem in an elegant way?

neubig:
@philip30 PTAL. I added `FeedbackLoss` here, which should probably work for your purposes?: https://github.com/neulab/xnmt/blob/refactor-loss/xnmt/loss_calculators.py#L161

philip30:
LGTM, so in order to use 'sampling' in the encoder, we have to wrap the loss using the FeedbackLoss. I'd try to fix my code, but probably starting tomorrow...

philip30:
Now I will fix the segmenting_encoder

philip30:
@neubig The translator is a ConditionedModel but not implementing the calc_loss function. I think this will be a problem in the future? calc_nll is the core method of the Translator. Maybe this shouldn't be a ConditionedModel?

philip30:
The FeedbackLoss still accumulate the loss from the sample by add_factored_loss_expr() and I am still having the same problem.

philip30:
Hmm OK I think I got your idea. But this means that it is not possible to do z_normalization across all samples and batches. I think it's not very important to do that, I will sacrifice that functionality for this cleanness.

neubig:
> The translator is a ConditionedModel but not implementing the calc_loss function. I think this will be a problem in the future? calc_nll is the core method of the Translator. Maybe this shouldn't be a ConditionedModel?

Sorry, we should change ConditionedModel, etc. to also be `calc_nll` not `calc_loss`.

> The translator is a ConditionedModel but not implementing the calc_loss function. I think this will be a problem in the future? calc_nll is the core method of the Translator. Maybe this shouldn't be a ConditionedModel?

If this is important functionality, a solution for this in `FeedbackLoss`, we could first run `LossCalculator.calc_loss()` multiple times (=`repeat`), do z-normalization, then call `calc_additional_loss()` multiple times. If not, it's easier to keep things as-is.

xnmt/models/translators.py
```diff
@@ -92,6 +92,7 @@ class DefaultTranslator(AutoRegressiveTranslator, Serializable, Reportable, base
     decoder: A decoder
     inference: The default inference strategy used for this model
     search_strategy:
+    global_fertility: A parameter for global fertility weight. 0 for no computation.
```
neubig: Is this comment a mistake?

philip30: Oops, yes this is a mistake.

neubig:
@philip30 Are you still working on this now? Or are you mostly done?

philip30:
I am still working on refactoring the MinRisk and Reinforce. I want to give them unittest as well. But I think we can do that in the latter PR? It will take a while and I am a bit busy with my experiment right now...

neubig:
Actually, I thought MinRisk and Reinforce were more-or-less already fixed. Is there something that wasn't finished with them?

philip30:
I want to take out the search strategy from the translator and put it in them. Also some fixes on the search strategy regarding the states, masks, and log_prob that are needed for the calculation. More or less is about cleaning up the old mess.

philip30:
Actually I might be done already. Just no unittest.

neubig:
Sure.

philip30:
OK! PTAL

philip30:
Fixed #494

philip30:
Currently doing regression testing on tanaka corpus. Maybe we can merge after that?

philip30:
Sorry, I think this was not very OK. The loss seems going down but the output tokens are garbage. I think we need to debug either the translator or the search_strategy.

neubig:
Woah woah woah. This is a branch I started that still had "WIP" in the title. You should wait to merge until the people involved have checked and agreed. I was also still working on fixing some things.

I need to sleep now, but I'll continue taking a look tomorrow. Could you please revert the merged commit and re-open a new PR?

PR #490, opened 2018-07-30
Better import conventions pt 2: introduced 'models' and 'transducers' packages

msperber:
While at it, I tried re-organizing things a bit by adding 'models' and 'transducers' packages. This won't make a difference in terms of avoiding circular imports, but may be a better solution that makes it easy to find code while at the same time reducing verbosity a bit (e.g. we now can do ```from xnmt.transducers import recurrent``` as opposed to the previous ```from xnmt import recurrent_transducers```).

Tests are still failing, I'll fix them if this re-organization is welcomed and found useful.


msperber:
Additional info:

- The ```models``` package contains ```base```, ```classifiers```, ```language_models```, ```retrievers```, ```translators``` modules. 
- The ```transducers``` package contains ```base```, ```convolution```, ```fixed_size_att```, ```positional```, ```pyramidal```, ```recurrent```, ```residual```, ```self_attention```, ```transforms``` modules.

msperber:
Sure. What do you think about further subpackages, e.g. xnmt.train (regimen, task), xnmt.loss (calculator, expr), xnmt.eval (metric, task), xnmt.data (reader, sent, output, ..), xnmt.modelparts (attender, embedded, decoder, ...)?

msperber:
Let me know if further subpackages are wanted, otherwise this is ready (missing dynet args are already included).

philip30:
I think they are all sound very nice. Not sure if expr should belong to loss though..

neubig:
I'd suggest we hold off on further subpackages for now. We can always change later.

msperber:
I've just finished subpackages for ```train```, ```eval```, ```modelparts``` because these are very clear and allow solving some disambiguities such as ```modelparts.transforms``` vs ```models.transformer```, ```modelparts.scorer``` vs ```eval.metrics.score```, etc.
Should I push these also? I haven't done anything about data and loss subpackages because these don't seem very clearly defined.

msperber:
I've pushed it so you can take a look, but let me know if parts should be reverted. Once this is merged I can take a look at actually making imports more consistent.

neubig:
I'm sure it's fine. Go ahead and merge.

Issue #383, opened 2018-05-02
Feature request: saving current model at regular intervals


pmichel31415:
It would be nice to be able to save the current model in a temporary file every x number of iterations/every x minutes, or even better save the state of training (current batch, optimizer state, weights).

This would help restart training in case of an unexpected interruption (out of memory, crash,...)

neubig:
This can already be done with the `dev_every` option. This triggers an evaluation on the dev set, but I think that should be OK? Does this seem like a reasonable solution?

With regards to saving the state of training, that would be good but require enhancements to DyNet.

Issue #367, opened 2018-04-27
Refactor experiments / resume crashed trainings


msperber:
It would be nice to do the following:
- Make config files and saved model files completely compatible. They are currently almost but not completely identical: Config files are dictionaries of ```!Experiment```, whereas saved models are a single ```!Experiment```.
- Keep a state in every component (especially ```TrainingRegimen```, but possibly also preprocessing and evaluation) that is stored as part of the model whenever the model is saved. If at initialization time the state is not zero, we fast-forward to the specified state.
- write out not only the *best* DyNet parameters, but also the *last* parameters.

This would allow the following:
- Via the state we can resume an experiment from the point it was last saved (in case it crashed, or in case one wants to kill it and resume on a different machine, etc.)
- Resuming could simply be achieved by running ```xnmt /location/to/saved_model.mod``` (if the saved experiment had been completed, this would result in XNMT exiting without doing anything).
- The ExperimentSeries could be extended to components that do automatic hyperparameter optimization such as Bayesian Optimization etc.

A config file would probably look like this for a config file with a single experiment:

```
!SimpleExperiment
  preproc: ..
    state: ..
  train: ..
    state: .. # the state is usually not given in the config file, 
                 #but included when the model is written out
  evaluate: ..
  state: ..
```
  
Or for a series of experiments (which is more similar to the current config files which always contain a series of experiments):

```
!ExperimentSeries
  experiments:
  - !SimpleExperiment
      state: ..
  - !SimpleExperiment ..
```

I believe this would be relatively easy to do, the main thing I'm not sure about how to handle best is that we would no longer have experiment names so ```{EXP}``` may no longer work.

neubig:
Is this fixed now? I'm not sure...

msperber:
No, I think nothing has been done along these lines yet.

msperber:
Making config files and saved experiments compatible has been implemented by #491.

Some thoughts on what would need to be done to support resuming crashed experiments:
- make .mod and .data files parallel. Currently there is only one .mod file and several .data dirs because the assumption was that .mod files don't change during the experiment, while the DyNet weights in .data do change. If we keep a state in the experiment, .mod files will also change over the course of the experiment, so we should have one .mod file be assigned to exactly one .data dir.
[...]

PR #491, opened 2018-07-30
Named experiments and resume option

msperber:
This is a proof-of-concept for #367 / #383:

- config files can now be single experiments, lists of experiments, or (as before) dictionaries of experiments. In the first 2 cases, experiments have to have a name argument, and the third case the name argument is taken from the dictionary key.
- this allows us to call ```xnmt /path/to/saved-model.mod``` which will resume the previous experiment.
- ```Experiment``` now has a simple state that checks whether everything is finished or not. If resuming a finished experiment, nothing will happen, otherwise the whole experiment is carried out.
- a ```--resume``` flag is added that makes sure that ```{EXP_DIR}```, ```{EXP}``` etc. is set to the value that was previously used, rather than to the values corresponding to the current context. This makes sure that locations of model and log files are the same as before.

In order to resume crashed trainings etc., all that's missing is for the training regimen to keep track of its state in a similar way and resume rather than start from the beginning if appropriate. This is actually not super easy so I'll leave it for a future commit.

One question I had is whether it's possible to write out the state of the DyNet trainer? Otherwise resuming a training would not produce the same results as running it without interruption.

### DIFF
--- a/docs/experiment_config_files.rst
+++ b/docs/experiment_config_files.rst
@@ -151,7 +151,13 @@ Minimal
 Multiple experiments
 ~~~~~~~~~~~~~~~~~~~~
 
-.. literalinclude:: examples/03_multiple_exp.yaml
+.. literalinclude:: examples/03a_multiple_exp.yaml
+    :language: yaml
+
+.. literalinclude:: examples/03b_multiple_exp.yaml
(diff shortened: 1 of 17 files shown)


msperber:
Done, I'll merge.

PR #493, opened 2018-07-31
Subword consistency report and mixed minor changes

msperber:
This adds a ```SubwordConsistencyReporter``` that gives statistics on subword consistency: recovered OOVs, fantasized new words, etc. I'm using it for character targets, but it might also be useful with BPE and other subwords.

Also includes several minor changes, including better debugging output and code consistency.


neubig:
Hi, two comments:
1. I'm actually not sure what "subword consistency" means. Maybe you could make it a more descriptive or accurate term.
2. This is not specific to this commit, but it's pretty frequent that we have big commits with lots of tiny changes to the formatting that are not central to the actual new functionality implemented. I think that's OK (it'd be a pain to try to go back and split into two commits), but it does make it harder to review, as I have to scroll through pages and pages of irrelevant stuff before finding the stuff that I actually need to be reviewing. When you make a commit like this, would you mind telling which files/functions actually require review, so I can focus on those? Thanks!

msperber:
Sure, sorry about that!
1. Character- or subword-based models can produce words that are not in the training. This can either be a good thing (when we produce a correct word that would have been an OOV with a word-based model) or a bad thing (when we produce something that's not a correct word). The reporter prints some statistics on this, and is therefore called "subword consistency" reporter. I couldn't think of a better name.. I can change it if you have a better suggestion, or otherwise I can improve the description in the doc string a bit.
An example output is:
```
Size of subword vocab:                      55
Unique words in training corpus:            32140
Unique words in test reference:             3831
Unique words in test hypothesis:            4089
OOVs in test reference:                     480 (1.1996700907250506%)
OOVs in test hypothesis:                    1526 (3.6065418793722817%)
Hypothesis OOVs precision (sentence-match): 2.686762778505898%
Hypothesis OOVs recall (sentence-match):    8.541666666666666%


Listing:
...
```

2. Here's a list of changes apart from the new reporter, probably none of them requires a review:
- alphabetic sorting of imports in __init__.py so it's easier to see what's missing
- use ReportOnException also when doing inference
- use the ```from xnmt import module``` import style in ```modelparts``` package (#439)
- updated type annotations, doc strings, base interface in ```bridges.py```
- reporters use platform-independent ```os.path.join```
- attention reporter by default reports only on the first 100 sentences so that we can simply add it by default to eval steps without worrying about extensive runtime or disk usage.
- attention reporter has a name (file prefix) so that subclasses can extend it while avoiding overwriting files
- log task name with the "> Checkpoint" output.

neubig:
1. What about `OOVAccuracyReporter` or `OOVStatisticsReporter` or something?
2. Thanks! I actually wasn't so much asking as a summary of all the changes here (in this case I've already read all the code), but rather when you make future PRs it'd be nice if you could point out something like "XXX and YYY are the major changes, and they can be found in the file `xxx.py` and `yyy.py`. There are many other minor formatting changes that probably don't need review." I think you've done that in the past, and it's helpful, so I just wanted to note that it'd be nice if you continue to do that when you have big commits with lots of cosmetic changes like this one.

msperber:
Sure, sounds good. I've renamed it to OOVStatisticsReporter and also expanded the explanation a bit.

xnmt/reports.py
```diff
@@ -447,3 +456,96 @@ def on_end_inference(self):
...
+    if hyp_words_total == 0: raise ValueError("Found empty hypothesis")
+    sorted_hyp_oovs_matched = sorted(hyp_oovs_matched.items(), key=lambda x: x[1], reverse=True)
+    sorted_hyp_oovs_unmatched = sorted(hyp_oovs_unmatched.items(), key=lambda x: x[1], reverse=True)
+    sorted_ref_oovs_unmatched = sorted(ref_oovs_unmatched.items(), key=lambda x: x[1], reverse=True)
+    num_oovs_ref = sum(ref_words_oov.values())
+    num_oovs_hyp = sum(hyp_words_oov.values())
+    num_oovs_hyp_matched = sum(hyp_oovs_matched.values())
+    hyp_oov_prec = f"{num_oovs_hyp_matched/num_oovs_hyp*100:.2f}%" if num_oovs_hyp>0 else "n/a"
+    hyp_oov_rec = f"{num_oovs_hyp_matched/num_oovs_ref*100:.2f}%" if num_oovs_ref>0 else "n/a"
+    with open(os.path.join(self.report_path, "subword-consistency.txt"), "w") as fout:
```
neubig: These two lines should also be changed to reflect the new name.

msperber: Thanks, done.

PR #500, opened 2018-08-03
Cleanup events  and training task

msperber:
This cleans up events and training tasks which had some issues that prevented me from working on something else:

- Events consist of callers and handlers. Handlers are designed nicely, but callers always needed to be part of some object even though they trigger global events. This made it unclear where they should placed, made it necessary to import and subclass EventTrigger which led to some circular imports, and caused awkward lines such as ```self.tasks[0].model.set_train(True)``` even though the event is not specific to ```task[0]```. Instead, I moved all callers into a new ```event_triggers.py``` as module-level functions.
- ```end_inference``` is no longer an event, instead the reporters (for which this event was introduced) now have ```create_sent_report()``` and ```conclude_report()``` methods that replace what was previously somewhat poorly named ```create_report()``` and ```on_end_inference()```
- ```TrainingTask``` base class interface was out of date and is now updated


neubig:
LGTM. Merge when you can resolve the conflicts.

PR #504, opened 2018-08-03
fix classifier and add f-score evaluator

msperber:
Fixes a bug in the ```SequenceClassifier``` related to batched decoding, and adds F-Score evaluation


msperber:
Thanks, I'll reflect the first one and ignore the second one then for now :)

PR #509, opened 2018-08-06
fix models base interface

### LINKED ISSUE #507: Potential model base interface issue
https://github.com/neulab/xnmt/blob/e06711ba0c666a828fd4c38ca7f0ac186c7863d8/xnmt/models/base.py#L67

Does this and other interface definitions of calc_nll need to be updated? It seems the return type is now a DyNet expression and loss_calculator is no longer passed.

msperber:
Will fix #507.

Is ```get_primary_loss``` still needed? The only place it's currently used is here, not sure if that's on purpose or was forgotten: https://github.com/neulab/xnmt/blob/master/xnmt/eval/tasks.py#L89

### DIFF
--- a/xnmt/eval/tasks.py
+++ b/xnmt/eval/tasks.py
@@ -85,13 +85,10 @@ class LossEvalTask(EvalTask, Serializable):
 
     loss_stats = {k: v/ref_words_cnt for k, v in loss_val.items()}
 
-    try:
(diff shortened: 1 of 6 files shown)


neubig:
I think `get_primary_loss` is no longer needed, but @philip30 might be able to confirm.

philip30:
Yea I think it is no longer needed.

msperber:
Got it, this should be ready then. PTAL.

PR #510, opened 2018-08-07
Tensorboard integration

msperber:
This PR uses the tensorboardX package to write out training logs that can be read by tensorboard. We previously wrote out the same information via the structured logger and I've simply replaced using the structured logger here by using the tensorboardX writer. The structured logger still exists but is not used currently. Tensorboard outputs are organized as ```exp_name/train_task_name/[train|dev]/[loss_name]```

Tensorboard outputs are written without further configuration necessary The only thing that has to be done to visualize outputs is to install tensorflow, and run ```tensorboard --logdir [logdir]```.

### DIFF
--- a/requirements-extra.txt
+++ b/requirements-extra.txt
@@ -1,3 +1,4 @@
-librosa
+librosa # for audio feature extraction
 sentencepiece>=0.0.6
 docopt
+tensorflow # for tensorboard visualization
--- a/requirements-extra.txt
+++ b/requirements-extra.txt
@@ -1,4 +1,0 @@
-librosa # for audio feature extraction
-sentencepiece>=0.0.6
-docopt
-tensorflow # for tensorboard visualization
--- a/requirements.txt
+++ b/requirements.txt
@@ -10,3 +10,4 @@ numpy
 Unidecode>=1.0.22
 beautifulsoup4
 nltk
+tensorboardX
--- a/docs/experiment_config_files.rst
+++ b/docs/experiment_config_files.rst
@@ -1,7 +1,10 @@
 .. _sec-exp-conf:
 
 Experiment configuration file format
-------------------------------------
+====================================
+
+Intro
+-----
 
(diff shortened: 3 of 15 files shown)


requirements-extra.txt
```diff
@@ -1,3 +1,4 @@
-librosa
+librosa # for audio feature extraction
 sentencepiece>=0.0.6
 docopt
+tensorflow # for tensorboard visualization
```
neubig: Ugh, I don't want to install all of tensorflow just to use tensorboard. I know this is the "extra" file, but people will need to install this for things like sentencepiece, which a large portion of the people using xnmt will need to install. Maybe we can add a brief document explaining how to do tensorboard visualization, which says at the beginning that you need to install tensorflow (if it's indeed necessary).

requirements.txt
```diff
@@ -10,3 +10,4 @@ numpy
 Unidecode>=1.0.22
 beautifulsoup4
 nltk
+tensorboardX
```
neubig: Actually, could we move `tensorboardX` to extra, and only import it when doing logging? It seems like it shouldn't be necessary unless we are actually doing tensorboard logging.

msperber: tensorboard logging is currently done by default and there's no option to disable it. We could add an option, but I thought that would only make things more complicated, and people might decide to want to look at the visualization after they already started training. It's also not expensive to do the logging. What do you think?

neubig: OK, that's fine I guess. I'm just wary of introducing more and more dependencies, which will break at some point when the package version changes, etc.

That being said, Tensorboard is a feature that I think a lot of people would want, so in this case it's probably fine.

requirements-extra.txt
msperber: Okay, I've reorganized these into a requirements-extra *directory* which contains separate files for the different extra use cases that can be installed via ```pip install -r ..```.

msperber:
I reorganized the extra requirements and also added a short doc page about visualizations (tensorboard and reporters). PTAL.

neubig:
Thanks!

PR #516, opened 2018-08-13
Revert bugfix and add more unit tests

msperber:
In #508 I had attempted to fix an issue that required every ```__init__``` argument of a ```Serializable``` to be stored via ```self.arg_name = arg_name```. This is done anyways pretty much everywhere in XNMT, but would be nice to get rid of. The fix had introduced some new issues though which are probably worse and can lead to model saving or loading failing. I haven't been able to come up with a fix yet, so this PR reverts the previous fix and adds several unit tests and minor refactoring that will help with debugging and developing a fix. Also included is a test config with a cascading example (which was the one that led to some problems and made me aware of this).

### DIFF
--- a/test/config/cascade.yaml
+++ b/test/config/cascade.yaml
@@ -35,8 +35,8 @@ cascade: !Experiment
     - !SimpleTrainingTask
       kwargs:
         << : *task1
-        src_file: 'examples/LDC94S13A.char'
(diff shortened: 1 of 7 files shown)


neubig:
I'm not familiar enough with the persistence code to 100% validate this, but I'll trust you on this!

Issue #495, opened 2018-08-02
Error Reports are not very helpful


philip30:
Let's maybe not printing the graph on the console as it can easily fill up your console and hard to scroll it up. Maybe we can dump it in a log, but not on the screen?

Also, let's maybe print the source and target using the real words not the ID. Using ID's are not very helpful.

philip30:
How about making error logger?

```python
err = logging.getLogging("error")

# In other files
from xnmt import err

.....
err.info(print_graph())
```

msperber:
We already have a file_logger that prints only to the log file, so you could try something like this:
```
def log_graph(): file_logger.error(dy.print_text_graphviz()); return "DyNet graph logged to file"
with utils.ReportOnException({"src": src, "trg": trg, "graph": log_graph}):
```

For printing actual words, I think overwriting ```ReadableSentence.__str__()``` would be the best way.

neubig:
I agree that this would be nice to fix. The computation graph is helpful sometimes, but it gets in the way the 95% of the time it's not helpful.

msperber:
fixed by #517.

PR #517, opened 2018-08-14
print CG on error only in debug mode

msperber:
My attempt to print it only to file but not stdout was not successful because the DyNet function uses print instead of returning a string, so I changed it to print the CG only with debug settings (solves part of #495).


msperber:
Update: also implemented a proper string representation for sentences, so this should solve #495 completely.

neubig:
Thanks!

PR #519, opened 2018-08-17
[WIP] LRU vocab for learning vocabularies on the flight using RL

philip30:
This is the implementation of the title.
Still one missing component. I don't know how to use the initializer to properly reset the value of the parameter given a row's number.  (line 190 and 268).

In the dynet documentation, so far, the initializer is just being passed around, with the real initialization code is invisible.
Maybe @neubig knows how to use the initializer?


philip30:
I tried these:

```python3
276     W_np = W.npvalue()
277     W_np[:, wordid] = 0
278     #W_np[:,wordid] = np.random.uniform(low=-1, high=1, size=self.hidden_dim)
279     print(W_np)
280     W.set_value(W_np)
281     print(W.dim())
282     print(W.npvalue())
283     print("--------------")
```

It outputs this:
```
[[ 0.         -0.76931858 -0.1804949   0.49332654 -0.13314922 -0.4454782 ]
 [ 0.          0.06065572  0.11964943  0.71070826  0.56009728 -0.50990784]
 [ 0.         -0.75339186 -0.13932016  0.69132841 -0.20233832  0.04961658]
 [ 0.          0.20901561  0.07602132 -0.58794713  0.26320916 -0.22998615]]
((4, 6), 1)
[[ 0.09559879 -0.76931858 -0.1804949   0.49332654 -0.13314922 -0.4454782 ]
 [-0.28111774  0.06065572  0.11964943  0.71070826  0.56009728 -0.50990784]
 [-0.14402415 -0.75339186 -0.13932016  0.69132841 -0.20233832  0.04961658]
 [-0.70089042  0.20901561  0.07602132 -0.58794713  0.26320916 -0.22998615]]
--------------
```

I think this is a bug of set_value? @neubig

pmichel31415:
@philip30 can you open an issue on d'une about this? I think I know how to fix this but I don't have the time right now.

To make your snippet work you should call `dy.renew_cg ()` after `set_value`

neubig:
Thanks @pmichel31415! @philip30 I think d'une is autocorrect for "Dynet" :)

philip30:
Ah, I see. So the problem is, I don't want to reset the pointer because this is done while training, not initializing. (So, I'd like to reset the parameter value while training)... If I don't call dy.renew_cg, would it become a problem? If yes, then it is currently not possible to do that.

xnmt/hyper_params.py
```diff
@@ -72,3 +72,8 @@ def __init__(self, sequence=None, initial=0.0, update=0):
   def update_value(self):
     return self.sequence[min(len(self.sequence)-1, self.update)]
 
+  def __hash__(self):
```
msperber: In #523 I had also been thinking about this, but learned that hash function must always return the same value. Hyper params are mutable, so I think we shouldn't implement a hash function here. Is there a particular reason you need this?

See also: https://hynek.me/articles/hashes-and-equality/

philip30: Actually, this was a quick hack to make my experiment running. Previously I have problem when serializing the hyperparams. Some code in the serialization tried to add the HyperParam to set() and leads to unhashable error. Currently I am still facing the serialization problem but other one. I will make a unittest and let you know if I can't fix that.

msperber: Okay yeah, I had encountered the same problem but it should already be fixed by #523.

msperber:
Sure, I'll take a look, though probably not before the weekend. The main source of errors has been the ```!Ref``` mechanism so that would be a starting point if you want to look into it before I do.

philip30:
@msperber OK I found the problem. The problem was because I was trying to save a python native object. The errors appears on this line 

```self.save_processed_arg("lrucache", self.lrucache)```

So there are two issues:
1. Serializer goes for an infinite recursive call when trying to save custom objects.
2. How to save the state of a custom object?

**Comment by philip30** [comment]
I tried to break down the data structure to python elementary data structure and I found that save_processed_arg can't be used for dictionary. I think this is a bug?

**Comment by msperber** [comment]
It depends on what kind of custom object it is. With normal, pickle-able objects it usually works fine (for example I've had no problem adding numpy arrays). There are some objects that are not serializable, mainly the DyNet (or other Cython) objects. Could you check if you can pickle this object using the standard Python features? My work-around for these things so far has been to introduce wrapper objects, e.g. the ones under ```optimizers.py```

**Comment by philip30** [comment]
Yes, I am trying what you are doing now. BTW, I pushed the new one. 

So now I will have an error when line 159 is written (and no error if it is not written):

```
158       print(type(self.cache_word_table)) # dict
159       self.save_processed_arg("cache_word_table", self.cache_word_table)
```

This is the traceback:
```
Traceback (most recent call last):
  File "/home/is/philip-a/.pyenv/versions/3.6.2/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/home/is/philip-a/.pyenv/versions/3.6.2/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/home/is/philip-a/tools/xnmt/xnmt/xnmt_run_experiments.py", line 123, in <module>
    sys.exit(main())
  File "/home/is/philip-a/tools/xnmt/xnmt/xnmt_run_experiments.py", line 108, in main
    raise e
  File "/home/is/philip-a/tools/xnmt/xnmt/xnmt_run_experiments.py", line 102, in main
    eval_scores = experiment(save_fct = lambda: save_to_file(model_file, experiment))
  File "/home/is/philip-a/tools/xnmt/xnmt/experiments.py", line 115, in __call__
    self.train.run_training(save_fct = save_fct)
  File "/home/is/philip-a/tools/xnmt/xnmt/train/regimens.py", line 155, in run_training
    self.checkpoint_and_save(save_fct)
  File "/home/is/philip-a/tools/xnmt/xnmt/train/regimens.py", line 161, in checkpoint_and_save
    save_fct()
  File "/home/is/philip-a/tools/xnmt/xnmt/xnmt_run_experiments.py", line 102, in <lambda>
    eval_scores = experiment(save_fct = lambda: save_to_file(model_file, experiment))
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 1395, in save_to_file
    f.write(_dump(mod))
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 1378, in _dump
    _resolve_serialize_refs(ser_obj)
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 1333, in _resolve_serialize_refs
    for _, node in _traverse_serializable(root):
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 647, in _traverse_serializable
    yield from _traverse_serializable(child, path_to_node.append(child_name))
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 647, in _traverse_serializable
    yield from _traverse_serializable(child, path_to_node.append(child_name))
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 647, in _traverse_serializable
    yield from _traverse_serializable(child, path_to_node.append(child_name))
  [Previous line repeated 1 more times]
  File "/home/is/philip-a/tools/xnmt/xnmt/persistence.py", line 344, in append
    raise ValueError(f"'{link}' is not a valid link")
ValueError: '0' is not a valid link
```

**Comment by msperber** [comment]
Okay, I think I understand what's going on now. The problem is that we're saving a dictionary with integer keys. This breaks some assumption in the persistence code, namely that paths to any place in the hierarchy can always be written using strings. If we have dictionaries that contain both ```'0'``` (the string) and ```0``` (the integer), paths would become ambiguous, so this is not something that can be fixed easily. Maybe there is any work-around you could think of? If you want to make changes to the persistence code I'll be happy to give hints, but don't have much time to work on this myself right now.

**Comment by msperber** [comment]
As an example why this is problematic: We could write something like
```
!LoadSerialized
  filename: ...
  path: model.encoder.segment_composer
  overwrite:
  - path: cache_word_table.0
     val: new_val
```

In this case, the ```.0``` normally gets resolved to an integer index if the parent object is a list, and a string index otherwise. If we permit arbitrary dictionary keys, this would no longer be possible to resolve unambiguously.

philip30:
I see. I think the first design was already good, but we just need to narrow down the assumption.
I guess my current code will work if I use a string based integer, right?

The second option would be always turning int to str when serializing via ```str()``` function?
Having both 0 and '0' is kind of rare design I guess, and we probably should not support the serialization of a too arbitrary dictionary. (The dictionary's key should be uni-type.)

msperber:
If you can use strings as dictionary keys, that should solve the problem, yes. And yeah, allowing uni-type dictionary keys would be a nice option, it probably wouldn't be too hard to implement (mainly the name_children and get_child methods would need to be changed, plus maybe the argument type checks), but would require a bit of testing to make sure nothing gets broken.

philip30:
These changes affect only my code, so with this push, nothing should break the main code. I am merging right now.

PR #523, opened 2018-08-20
Hyperparams as numbers.Real

msperber:
Using ```hyperparam.Scalar``` etc. in my config file led to a failing type check because ```Scalar``` is not a ```float```.

To solve this, I made ```Scalar``` an instance of Python's abstract base class ```numbers.Real```, and updated the type annotations to use numbers.Real instead of float. Also includes some general improvements toward more consistent type annotations.

I think as a general goal we should try to use the abstract base class type variants whenever possible (e.g. also for ```numbers.Integral```, and possibly for a new string-type ABC) because that will enable making hyperparams and ```!Ref``` to be compatible with these types.

Update: I also slightly updated ```DefinedSequence``` to use its first list item as initial value.

### DIFF
--- a/xnmt/batchers.py
+++ b/xnmt/batchers.py
@@ -4,6 +4,7 @@ import math
 import random
 from abc import ABC, abstractmethod
 from functools import lru_cache
+import numbers
 
 import numpy as np
 import dynet as dy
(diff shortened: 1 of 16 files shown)


msperber:
Sorry for the premature commit, but this should be ready now. I ran into trouble because ```DefinedSequence``` is mutable and therefore not hashable, and some bugchecking-code in persistence.py assumed that all serializable objects are hashable, but now no longer makes this assumption.

msperber:
The main file to review would be ```hyper_params.py```. If this is fine, everything else should be, too.

philip30:
LGTM, but why one of the test is failing?

msperber:
Seems it was a Travis issue, restarting the test fixed it. I'll merge, thanks for taking a look!

PR #527, opened 2018-08-24
fix MLP

msperber:
The MLP code was not calling add_serializable_component properly, which is fixed in this PR. I don't think it was actually used in any of the example configs so it's not a critical bug, but should be fixed anyways.


msperber:
This doesn't need a review, I'll merge right away when tests are passing.

PR #528, opened 2018-08-24
Sentence Filter after corpus reading

gmwe:
Added the ability to apply sentence filters after reading the corpus (Not just as a preprocessing step).
example usage:
```
train: !SimpleTrainingRegimen
  sent_filter: !SentenceFiltererMatchingRegex
    spec:
      regex_trg: '^[^\(\)]*$'
```


msperber:
Looks good to me, but it seems unit tests are currently failing?

neubig:
Should sentence filtering be under "train" directly? It seems like maybe it should be a function of the readers?

Issue #530, opened 2018-09-03
Decoding of input line-by-line not supported


neubig:
Sometimes we want to set up an MT system and have it translate input one at a time (e.g. when actually deploying MT systems). This is not possible at the moment using our current setting.

It is possible to translate from stdin:

    xnmt_decode --src /dev/stdin --hyp /dev/stdout --mod examples/models/standard.mod

But it will not translate until it gets the "EOF" sign indicating the end of the file and the stream closes.

I'll work on fixing this.

Issue #526, opened 2018-08-22
Batch decoding


neubig:
Currently xnmt decodes one sentence at a time, but it would be nice to be able to do multiple sentences at the same time to take advantage of batching.

PR #531, opened 2018-09-03
Created stream decoding

neubig:
Fixes https://github.com/neulab/xnmt/issues/530

Note that this results in two alternative code paths for decoding: one for "batch" where the entire file is read in, batched, and processed, and one for "stream" where a single example is decoded at a time. Currently I set "stream" to default, as batch decoding isn't implemented in a way that makes it any more efficient (see https://github.com/neulab/xnmt/issues/526).

### DIFF
--- a/xnmt/inferences.py
+++ b/xnmt/inferences.py
@@ -107,6 +107,7 @@ class Inference(object):
     """
     Generate outputs for a single batch and write them to the output file.
     """
+    batch_size = src_batch.batch_size()
     src_len = src_batch.sent_len()
     if max_src_len is not None and src_len > max_src_len:
       output_txt = "\n".join([NO_DECODING_ATTEMPTED] * batch_size)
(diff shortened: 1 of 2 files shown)


neubig:
Ahh, so peek at the batch size and read in that many sentences every time? Sounds like a good idea, I'll do this. Maybe also add a check that we're using InOrderBatcher.

neubig:
@msperber PTAL and merge if it looks good.

xnmt/inferences.py
```diff
@@ -32,13 +33,15 @@ class Inference(object):
             * ``score``: output scores, useful for rescoring
     batcher: inference batcher, needed e.g. in connection with ``pad_src_token_to_multiple``
     reporter: a reporter to create reports for each decoded sentence
+    stream: whether to perform streaming inference (one input at a time) from a file
```
msperber: Is this still needed?

neubig: No it wasn't, thanks for the catch! Removed.

PR #537, opened 2018-10-10
Fix rolling avg

msperber:
The rolling average is used to skip updates for noisy gradients but has caused occasional problems, maybe because of numerical stability issues. This PR implements a workaround.


philip30:
It is currently not trivial to judge what's going on by looking at the warn() message. Can you elaborate more?

msperber:
I'm honestly not quite sure ;-) I doubled-checked the math and can't find a mistake there, so numerical problems is the only explanation I have right now. It did happen several times, and this fix helped to prevent the crash and led to reasonably converged models.

PR #541, opened 2018-11-01
Fixed Reporting

philip30:
I fixed reporting so that compute_report is determined if there is a reporter being defined, rather than making a separate flag in exp_global. The fix includes if there are multiples eval_tasks that want to write reports to different places. 

Having compute_report on exp_global is somewhat confusing, and the report should just be computed as needed (if the Reporter is defined).


philip30:
+1 metric to calculate segmentation F-measure.

philip30:
Any comment @msperber?

msperber:
I've been a bit busy, but I will take a look within the next days!

xnmt/eval/tasks.py
```diff
@@ -110,6 +110,8 @@ class AccuracyEvalTask(EvalTask, reports.Reportable, Serializable):
     model: generator model to generate hypothesis with
     eval_metrics: list of evaluation metrics (list of Evaluator objects or string of comma-separated shortcuts)
     inference: inference object
+    perform_inference: Whether to generate the output or not. One eval task can use difference hyp_file that is generated
```
msperber: ``difference hyp_file that is generated`` -> ``an already existing hyp_file that was generated`` ?

xnmt/inferences.py
```diff
@@ -150,6 +150,7 @@ def _generate_output(self, generator: 'models.GeneratorModel', src_file: str,
     with open(trg_file, 'wt', encoding='utf-8') as fp:  # Saving the translated output to a trg file
       cur_sent_i = 0
       src_batch, ref_batch, assert_batch = [], [], []
+      event_trigger.set_reporting(self.reporter is not None)
```
msperber: How about also calling ``event_trigger.set_reporting(False)`` after inference is finished? That way, we don't have to worry about e.g. having a reporter in a dev checkpoint that would then potentially cause training to slow down.

xnmt/eval/tasks.py
philip30: Done!

xnmt/inferences.py
philip30: Nice idea. Implemented.

philip30:
I have reflected all the comments, now I am merging. Thanks, @msperber !

PR #542, opened 2018-11-12
Splitting one big file into module

philip30:
Hi guys,

XNMT is getting bigger and it is hard to add one more class on top of the existing one in the same file. The usual workaround is to split the big file into modules, giving one class a single file within the module. I have some samples of this refactoring done on the input_readers and translators. If this looks good, I can refactor more files.

The goodness of this structure is that it is easy to locate the source code of a particular class (unlike before). The downside is that there will be a bunch of files!


msperber:
Hmm currently the rough idea is to have in each file a base class plus its subclasses. Subclasses are often quite similar (e.g. NpzReader and H5Reader are almost the same), and having them in the same file makes it easy to see what is similar, what code can be shared etc. My feeling is that splitting into a lot of different files would make things more complicated, also because the imports will become more complex. And most IDEs make it easy to find any particular class by just searching for the name?

I had also been thinking of making a distinction between "core" and "extra" features to keep things a bit more organized, but it seems difficult to know where to draw the line (although we have that in one place already, the ```specialized_encoders``` package).

philip30:
If you see this PR more closely, the importing is unchanged to the outside
package. The trick is to made some classes available for import in the
module level, (importinv from init), and later those classes can be
imported from the module, not the file.

philip30:
Any comment from @neubig ?

neubig:
Honestly, I don't have a strong preference either way.

philip30:
I think I will try to do it on input readers and translator first (Since there are really many types of inputs) and the file is really getting big. If we have the problem in the future with this design, I'd be happy to refactor it back.

To be honest, having a 'specialized' folder is actually against the spirit of the XNMT. We should be able to design any type of encoder we want, without having to be marked 'too-specialized', as long as it works with certain input types and work with other components. Just, the current way of organizing everything is a bit messy I guess.

We can, for example, group the files that share the same code on the same folder, which is also easy to be compared?

neubig:
I'm very happy to have you refactor to make things cleaner! One of the reasons why we made the "specialized encoders" thing was because there were a few encoders that were implemented in a somewhat messy and specialized way (e.g. a 3-layer convolutional encoder with fixed size hidden sizes that wasn't parameterized to make this adjustable). I didn't want these to be part of the main xnmt code base, because they should at least be appropriately configurable.

msperber:
> We can, for example, group the files that share the same code on the same folder, which is also easy to be compared?

I also think it's fine to split files and make them smaller when they get too big, I can see that for the files that you mentioned! But I don't think we should necessarily go for the one-class-per file extreme, because it's not very Pythonic, at least I haven't seen that in other big Python projects so far.

philip30:
currently closing this, I will refactor on a different head base.

PR #553, opened 2018-11-22
Autobatch example

msperber:
The tiny example is moved to the unit tests, the big ones are merged into one config file and with removed default arguments. @beckdaniel could you check if this is as intended?


beckdaniel:
It looks ok to me. Maybe a test with a expected failure when calling the autobatching regimen with `batch_size != 1`?

msperber:
Sounds good, it's done. I'll merge!

Issue #563, opened 2019-01-13
Overabundance of forced decoding


armatthews:
Right now every search strategy has special logic to handle forced decoding, but the results of forced decoding should be the same no matter what search strategy one uses.

I propose that we eliminate forced decoding logic from existing search strategies and implement a special forced decoder class that implements this functionality.

Issue #561, opened 2019-01-09
Search Strategies assume decoder supports calc_log_probs


armatthews:
Right now all search strategies rely on the decoder being able to provide (log) probabilities over the whole vocabulary. For some setups this might be woefully inefficient (e.g. class-factored softmaxes), and for others it might be downright impossible (e.g. my open-vocabulary morphology [work](http://www.aclweb.org/anthology/N18-1130)).

I propose we refactor things so that decoders implement operators ``top_k`` and/or ``sample``. These functions should return (word, score) pairs. I believe this change would maintain the functionality of all existing decoders and search strategies, while allowing for a much broader class of decoders to be implemented.

Thoughts?

neubig:
I think the concept like this is already implemented in the Scorer class
somewhat. Probably the decoders would need to be modified to use the
functions implemented in Scorer appropriately.

neubig:
Sorry, not the Decoder class, but whether the SearchStrategy class.

PR #565, opened 2019-01-17
Refactoring decoders/search_strategies

armatthews:
This refactors decoders and search_strategies to use new interface calls best_k() and/or sample() instead of relying on the existence of a decoder.calc_log_probs() function.

For certain types of decoders computing the whole (log) distribution is computationally difficult or down right impossible. For example, hierarchical softmax, my morphological stuff, or the RNNG decoder with it's structured output actions.


armatthews:
This also addresses issues [561](https://github.com/neulab/xnmt/issues/561) and [563](https://github.com/neulab/xnmt/issues/563).

philip30:
Hey @armatthews I am not sure if we can also remove the attention_vecs from the attender? I think I was the one who added that and the purpose is to get the last attention score or be used somewhere. As you said, it is better to make these components stateless.

PR #569, opened 2019-02-27
Stative Attenders

armatthews:
This PR enables stative attenders, and contains a sample implementation of "Modeling Coverage for Neural Machine Translation" ([Tu et al. 2016](http://www.aclweb.org/anthology/P16-1008)).

### DIFF
--- a/xnmt/modelparts/attenders.py
+++ b/xnmt/modelparts/attenders.py
@@ -8,37 +8,45 @@ from xnmt import logger
 from xnmt import batchers, expression_seqs, events, param_collections, param_initializers
 from xnmt.persistence import serializable_init, Serializable, Ref, bare
 
+class AttenderState(object):
+  pass
+
 class Attender(object):
   """
   A template class for functions implementing attention.
   """
 
-  def init_sent(self, sent: expression_seqs.ExpressionSequence) -> None:
+  def init_sent(self, sent: expression_seqs.ExpressionSequence) -> AttenderState:
     """Args:
          sent: the encoder states, aka keys and values. Usually but not necessarily an :class:`expression_seqs.ExpressionSequence`
     """
     raise NotImplementedError('init_sent must be implemented for Attender subclasses')
 
-  def calc_attention(self, state: dy.Expression) -> dy.Expression:
+  def calc_attention(self, dec_state: dy.Expression, att_state: AttenderState = None) -> dy.Expression:
     """ Compute attention weights.
 
     Args:
-      state: the current decoder state, aka query, for which to compute the weights.
+      dec_state: the current decoder state, aka query, for which to compute the weights.
+      att_state: the current attender state
     Returns:
       DyNet expression containing normalized attention scores
     """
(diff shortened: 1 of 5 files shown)


xnmt/modelparts/attenders.py
```diff
@@ -6,47 +6,66 @@
...
 class Attender(object):
   """
   A template class for functions implementing attention.
   """
 
-  def init_sent(self, sent: expression_seqs.ExpressionSequence) -> None:
+  def init_sent(self, sent: expression_seqs.ExpressionSequence) -> AttenderState:
     """Args:
          sent: the encoder states, aka keys and values. Usually but not necessarily an :class:`expression_seqs.ExpressionSequence`
     """
```
msperber: Return value needs documentation.

xnmt/modelparts/attenders.py
```diff
@@ -55,56 +74,133 @@ class MlpAttender(Attender, Serializable):
...
+    self.coverage_updater = self.add_serializable_component(
+        "coverage_updater",
+        coverage_updater,
+        lambda: recurrent.UniGRUSeqTransducer(
+            input_dim=(self.state_dim + self.hidden_dim + 1),
+            hidden_dim=self.coverage_dim,
+            param_init=param_init,
+            bias_init=bias_init))
+
   def init_sent(self, sent: expression_seqs.ExpressionSequence) -> None:
```
msperber: Return type outdated.

xnmt/modelparts/attenders.py
```diff
@@ -6,47 +6,66 @@
...
+      attention: the attention vector to use. if not given it is calculated from the state(s).
     """
-    attention = attention or self.calc_attention(state)
+    attention = attention or self.calc_attention(dec_state, att_state)
     I = self.curr_sent.as_tensor()
     return I * attention
 
+  def update(self, dec_state: decoders.DecoderState, att_state: AttenderState, attention: dy.Expression):
+    return None
+
```
msperber: Could you document how ```update```'s intended use is?

armatthews:
Thanks for the feedback, Matthias! I updated the documentation.

I talked to Graham about the mechanism of compute_attention() and update(). I agree that it would be nice to factor this so the translator class doesn't have to call update() but I'm not sure that's possible in general.

The real problem is that the attention vector returned by the attender may not be the "final" attention vector used downstream. For example, if one chooses to ensemble multiple attenders then the final attention vector will not be the same as the vector produced by any individual one. This mechanism allows for us to feed the real attention vector back into the attender even in these types of cases.

I talked to Graham a bit about this, and this was the best solution we came up with. I'm happy to discuss further if you see a better way!

msperber:
I see, yeah I had suspected something like that. In that case I think this can be merged (once the merge conflicts are resolved)!

-- Matthias
_(this comment is "not a contribution")_

neubig:
@armatthews If you resolve the conflicts on this I think we can merge.

